PR #18, opened 2019-05-17
WASM version

AndriusA:
Basic WASM module to create a blocking engine and check urls against network rules.

From benchmarks appears to be slower (more than 40%) than just Rust-neon based node native wrapper


bbondy:
Maybe we should just keep this in a branch to avoid people picking it by accident until we have a use case for it? The only use case I knew was in case it was faster than the node wrapper.

AndriusA:
Agreed we should keep it in a branch until decided otherwise - could be a useful execution target for some scenarios, but not needed for us

snyderp:
One suggested use case is to inject into pages as a service worker on iOS, so that iOS can take advantage of (among other things) uBO style request proxying / rewriting

PR #26, opened 2019-06-05
changes tag retention behaviour

AndriusA:
retain tags of current instance, do not serialize.

when deserializing, tags of the existing instance (the one we are recreating lists into) are maintained

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -47,6 +47,7 @@ impl Engine {
     }
 
     pub fn deserialize(&mut self, serialized: &[u8]) -> Result<(), BlockerError> {
+        let current_tags = self.blocker.tags_enabled();
         let gz = GzDecoder::new(serialized);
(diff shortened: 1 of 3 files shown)


src/engine.rs
```diff
@@ -155,15 +157,16 @@ mod tests {
...
+            ("https://brianbondy.com/about", false),
+            ("https://brave.com/about", false),
         ];
 
         let mut engine = Engine::from_rules(&filters);
         engine.tags_enable(&["stuff"]);
         engine.tags_enable(&["brian"]);
         let serialized = engine.serialize().unwrap();
         let mut deserialized_engine = Engine::from_rules(&[]);
+        deserialized_engine.tags_enable(&["stuff"]);
```
AndriusA: The new engine we'd be restoring to needs to set its own tags

PR #29, opened 2019-06-06
Implements uBO style polyfills for requests redirects

### LINKED ISSUE #27: Handle uBO style polyfills for blocked requests
uBO defines polyfills, or alternate content bodies that should be returned for resources from a URL that matches a pattern.  The bodies of these responses, and identifiers for them, are given here:

https://github.com/uBlockOrigin/uAssets/blob/master/filters/resources.txt

Rules then reference these polyfils with the `redirect=<identifier>` option on rules.

It would be good if adblock-rust both could 
1) understand the redirect option in rules
2) parse polyfil definitions like https://github.com/uBlockOrigin/uAssets/blob/master/filters/resources.txt 
3) and return the alternate message body when appropriate

AndriusA:
returns request content to use for requests matched on rules with `redirect` option

Closes #27

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -117,6 +117,9 @@ impl Blocker {
                         Some(data_url.trim().to_owned())
                     } else {
                         // TOOD: handle error - throw?
+                        if self.debug {
+                            eprintln!("Matched rule with redirect option but did not find corresponding resource to send");
+                        }
                         None
                     }
                 } else {
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -43,7 +43,6 @@ impl From<FilterError> for BlockerError {
     }
 }
 
-
 #[derive(Serialize, Deserialize)]
 pub struct Blocker {
     csp: NetworkFilterList,
@@ -64,7 +63,7 @@ pub struct Blocker {
     load_cosmetic_filters: bool,
     load_network_filters: bool,
 
-    #[serde(skip_deserializing)]
+    #[serde(default)]
     resources: Resources
 }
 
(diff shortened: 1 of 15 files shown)


AndriusA:
@snyderp if you review `tests/ublock-coverage.rs` and the tests in `src/resources.rs` it should be clear what gets handled - please check it is all as intended.

Also added a test that checks parsing extracts all 110 resources for the snapshot of `resources.txt` included in the repo

src/blocker.rs
```diff
@@ -95,11 +100,28 @@ impl Blocker {
...
+                    if let Some(resource) = blocker_redirects.get_resource(redirect) {
+                        let mut data_url: String;
+                        if resource.content_type.contains(';') {
+                            data_url = format!("data:{},{}", resource.content_type, resource.data);
+                        } else {
+                            data_url = format!("data:{};base64,{}", resource.content_type, base64::encode(&resource.data));
+                        }
+                        Some(data_url.trim().to_owned())
+                    } else {
+                        // TOOD: handle error - throw?
```
snyderp: Could this be logged somehow, even if its just mirroring the "didn't understand filter: X" stuff the current lib does?  Would be a nice, noisy reminder if there is some new filter format we don't support, something like that

snyderp:
Also, is this exposed through the node API?

AndriusA:
@snyderp should we merge and close? It is backwards-compatible, so don't see any harm in that, FFI and browser can be wired up in due course

snyderp:
Yep, looks good to me, though it looks like there are some conflicts now.  But I think the change is terrific!

AndriusA:
Merged master to resolve conflicts

Also to ensure deserialization is backwards compatible have to ignore deserializing of redirect resources for now - included the required changes, tagging @bbondy for review

AndriusA:
Switching to [MessagePack](https://msgpack.org) for serialization to have a cleaner way of adding new fields to blocker structures over time - this works as intended with `#serde(default)` field property by using default value for the field if it isn't found when deserializing.

PR #30, opened 2019-06-06
adds more detailed info to Node wrapper response

AndriusA:
Check the updated README for usage example - _both_ the client and the `check` call have an extra `debug = true` parameter:

- the one in the client retains string representation of rules
- the one in the `check` returns detailed information object rather than only `true` or `false` for a match

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,14 +43,18 @@ Note the Node.js module has overheads inherent to boundary crossing between JS a
 const AdBlockClient = require('adblock-rs');
 let rules = fs.readFileSync('./data/easylist.to/easylist/easylist.txt', { encoding: 'utf-8' }).split('\n');
 
-const client = new AdBlockClient.Engine(rules);
+// create client with debug = true
+const client = new AdBlockClient.Engine(rules, true);
 
 const serializedArrayBuffer = client.serialize(); // Serialize the engine to an ArrayBuffer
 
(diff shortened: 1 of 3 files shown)


AndriusA:
Node repl output as an example of use:

```
> const AdBlockClient = require('adblock-rs');
>
> let rules = fs.readFileSync('./data/easylist.to/easylist/easylist.txt', { encoding: 'utf-8' }).split('\n');
>
> const client = new AdBlockClient.Engine(rules, true);
>
> const serializedArrayBuffer = client.serialize(); // Serialize the engine to an ArrayBuffer
>
> console.log("Matching:", client.check("http://example.com/-advertisement-icon.", "http://example.com/helloworld", "image", true))
Matching: {
  matched: true,
  explicit_cancel: false,
  redirect: null,
  exception: null,
  filter: '-advertisement-icon. <+> /icon_ad. <+> /icon_ads_ <+> /icon_advertising_'
}
> console.log("Matching:", client.check("https://github.githubassets.com/assets/frameworks-64831a3d.js", "https://github.com/AndriusA", "script", true))
Matching: {
  matched: false,
  explicit_cancel: false,
  redirect: null,
  exception: null,
  filter: null
}
> console.log("Matching:", client.check("http://example.com/-advertisement-icon.", "http://example.com/helloworld", "image"))
Matching: true
>
```

PR #42, opened 2019-07-30
Add important field to BlockerResult

antonok-edm:
Currently, information about whether or not a filter is `important` is lost before it is returned as a `BlockerResult`. This adds an extra boolean field to `BlockerResult` for that purpose.

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -115,7 +115,7 @@ impl Blocker {
      * Decide if a network request (usually from WebRequest API) should be
      * blocked, redirected or allowed.
      */
-    pub fn check(&self, request: &Request) -> BlockerResult {
+    pub fn check(&self, request: &Request, skip_unimportant: bool, skip_exception: bool) -> BlockerResult {
         if !self.load_network_filters || !request.is_supported {
             return BlockerResult::default();
         }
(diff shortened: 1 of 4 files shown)


antonok-edm:
Intended as part of the fix for https://github.com/brave/brave-browser/issues/5440

antonok-edm:
@AndriusA I would bump the minor version number but I don't know your policy on that.

AndriusA:
The changes look good to me, but could you squash the commits into one for a cleaner log of what changed?

src/blocker.rs
```diff
@@ -155,54 +157,69 @@ impl Blocker {
...
-            // Set `bug` of request
-            if !f.is_important() {
-                #[cfg(feature = "metrics")]
-                print!("exceptions\t");
-                if f.has_bug() {
+        let exception = if skip_exception {
+            None
+        } else {
+            match filter.as_ref() {
+                None => self.exceptions.check(request, &request_tokens),
```
AndriusA: so in the default mode of operation, `skip_unimportant = false` and `skip_exception = false` this will still check all exceptions when no filter matches?

src/blocker.rs
```diff
@@ -111,7 +113,7 @@ impl Blocker {
      * Decide if a network request (usually from WebRequest API) should be
      * blocked, redirected or allowed.
      */
-    pub fn check(&self, request: &Request) -> BlockerResult {
+    pub fn check(&self, request: &Request, skip_unimportant: bool, skip_exception: bool) -> BlockerResult {
```
AndriusA: skip_unimportant appears to only be used below with redirects?

src/blocker.rs
antonok-edm: Yes, because exceptions are checked before other filters anyways.

antonok-edm: Would it make sense to skip exceptions after an important rule has been encountered, within the same engine? And likewise to skip normal rules if any important or exception rules were encountered? This would change some `BlockerResult` outputs.

AndriusA:
done some tweaks the way I see fit to undo some unnecessary changes and clarify parameters

PR #47, opened 2019-08-29
Implement cosmetic filtering

### LINKED ISSUE #38: Implement cosmetic blocking [tracking]
Currently, `adblock-rust` only deals with network filtering. There is no mechanism for parsing cosmetic rules and applying them in an efficient manner to pages as they load.

antonok-edm:
Closes #38.


src/engine.rs
```diff
@@ -126,25 +146,46 @@ impl Engine {
...
-                    e => eprintln!("Encountered unexpected error {:?} when checking for filter existence", e),
-                }
-                
-                self
-            }
+        let filter_parsed = parse_filter(filter, true, true, true);
+        match filter_parsed {
+            Ok(ParsedFilter::Network(filter)) => self.add_network_filter(filter),
+            Ok(ParsedFilter::Cosmetic(filter)) => self.add_cosmetic_filter(filter),
+            _ => {},
```
AndriusA: nit: when doing network filters, there was a wish to have some debug output in case provided input is bad, here cosmetic filter errors get swallowed completely

src/engine.rs
```diff
@@ -164,17 +205,29 @@ impl Engine {
...
     pub fn with_resources<'a>(&'a mut self, resources: &[Resource]) -> &'a mut Engine {
         self.blocker.with_resources(resources);
+        self.cosmetic_cache.use_resources(resources);
         self
     }
 
     pub fn resource_add<'a>(&'a mut self, resource: Resource) -> &'a mut Engine {
-        self.blocker.resource_add(resource);
+        self.blocker.resource_add(&resource);
+        self.cosmetic_cache.add_resource(&resource);
```
AndriusA: are we using the same instance of Engine in the end for network and cosmetic filtering? If not, this will unnecessarily increase memory use and should keep track of which one it is instead

src/filters/cosmetic.rs
```diff
@@ -0,0 +1,1493 @@
...
+            let colon_index = colon_index + index_after_colon;
+            index_after_colon = colon_index + 1;
+            if line[index_after_colon..].starts_with("style") {
+                if line.chars().nth(index_after_colon + 5) == Some('(') && line.chars().nth(line.len() - 1) == Some(')') {
+                    *selector = &line[suffix_start_index..colon_index];
+                    *style = Some(line[index_after_colon + 6..line.len()-1].to_string());
+                } else {
+                    return Err(CosmeticFilterError::InvalidStyleSpecifier);
+                }
+            } else if line[index_after_colon..].starts_with("-abp-")
```
AndriusA: not on the critical path and not checked what compiler maps it to, but could store `line[index_after_colon..]` in a variable and use that instead of repeating

src/filters/cosmetic.rs
```diff
@@ -0,0 +1,1493 @@
...
+            self.not_entities.is_some() ||
+            self.not_hostnames.is_some()
+    }
+
+    /// Check whether this rule applies to content from the hostname and domain corresponding to
+    /// the provided hash lists.
+    ///
+    /// See the `matches_str` test helper function for an example of how to convert hostnames and
+    /// domains into the appropriate hash lists.
+    pub fn matches(&self, request_entities: &[Hash], request_hostnames: &[Hash]) -> bool {
```
AndriusA: method makes sense, but not seeing where it gets used besides tests

src/cosmetic_filter_cache.rs
```diff
@@ -0,0 +1,680 @@
...
+        let mut rules_that_apply = vec![];
+        for hash in request_entities.iter().chain(request_hostnames.iter()) {
+            if let Some(specific_rules) = self.specific_rules.retrieve(hash) {
+                rules_that_apply.extend(specific_rules);
+            }
+        };
+
+        let mut exceptions = HostnameExceptionsBuilder::default();
+
+        rules_that_apply.iter().for_each(|r| {
```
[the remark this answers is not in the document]
antonok-edm: I guess `insert` is not the best name for this function, but it saves rules that are exceptions and ignores any other rules. I'll rename it to `insert_if_exception` and add a doc comment as well.

src/engine.rs
antonok-edm: I reused the main (i.e. not the regional or custom) Engine instance in my `brave-core` PR.

antonok-edm: Are resources ever added to the other Engines?

src/filters/cosmetic.rs
antonok-edm: This was from before I started storing rules keyed by hostname.

It might still be useful for some kind of logging or dashboard? Should I just remove it altogether?

AndriusA: let's remove it, if we want it again we can probably do better

src/engine.rs
AndriusA: ok, if reusing that's fine, will reduce the overhead when combining all filters to one engine instance.

But would that mean cosmetic filters don't apply for regional?

PR #50, opened 2019-09-26
Migrate to web_accessible_resources and scriptlets.js

antonok-edm:
Closes https://github.com/brave/brave-browser/issues/4564

With the deprecation of the `resources.txt` file, this allows `adblock-rust` to pull resources from uBlock Origin's newer `web_accessible_resources` directory and `scriptlets.js` file. Redirectable resources from `web_accessible_resources` and resources for cosmetic injection from `scriptlets.js` can both be represented as `Resource` structs.

Snippet from `src/resources.rs`:
``` rust
/// Struct representing a resource that can be used by an adblocking engine.
///
/// - `name`: Represents the primary name of the resource, often a filename
/// - `aliases`: Represents secondary names that can be used to access the resource
/// - `kind`: How to interpret the resource data within `content`
/// - `content`: The resource data, encoded using standard base64 configuration
pub struct Resource {
    pub name: String,
    pub aliases: Vec<String>,
    pub kind: ResourceType,
    pub content: String,
}

/// Different ways that the data within the `content` field of a `Resource` can be interpreted.
///
/// - `Mime(type)` - interpret the data according to the MIME type represented by `type`
/// - `Template` - interpret the data as a Javascript scriptlet template, with embedded template parameters in the form of `{{1}}`, `{{2}}`, etc.
pub enum ResourceType {
    Mime(MimeType),
    Template,
}

pub enum MimeType {
    ImageGif,
    TextHtml,
    ApplicationJavascript,
    AudioMp3,
    VideoMp4,
    ImagePng,
    TextPlain,
    Unknown,
}
```

The resulting `Resource`s can be loaded into the engine directly, or serialized to JSON for transfer to another instance. The equivalent data structures in JSON are as follows:

``` typescript
// Resource
{
    "name": String,
    "aliases": String[],
    "kind": ResourceType,
    "content": String,
}

// ResourceType
{ "mime": MimeType } | "template"

// MimeType
"image/gif" | "text/html" | "application/javascript" | ... | "application/octet-stream"
```

A new Javascript function `uBlockResources` is exposed through `adblock-rust`'s neon bindings. This function takes 3 string arguments, namely:
- The path to uBlock Origin's `web_accessible_resources` directory
- The path to uBlock Origin's `redirect-engine.js` file
- The path to uBlock Origin's `scriptlets.js` file

and will return a list of corresponding `Resource` JavaScript objects. In combination with a locally checked out version of the uBlock Origin source repo, this function can be used to produce a new set of resources for the adblocking engine.


src/redirect_resources.rs
```diff
@@ -0,0 +1,101 @@
...
+        });
+
+        Self {
+            resources: redirectable_resources,
+            aliases,
+        }
+    }
+
+    pub fn get_resource(&self, name: &str) -> Option<&RedirectResource> {
+        self.resources.get(name).or_else(|| self.aliases.get(name).and_then(|name| self.resources.get(name)))
```
AndriusA: IMO this would be better handled with ref-counted pointers to avoid additional copies and a two-step lookup

antonok-edm: My only reason for avoiding that was that `serde` doesn't really handle `Rc` correctly. The default implementation, which requires a `cfg` flag, ends up just cloning the data every time.

The primary name is generally the shortest one, so the second lookup should be relatively uncommon anyways.

It would be possible to get it working with some added logic (e.g. custom serializer / deserializer). Would you say it's worth the additional complexity?

PR #58, opened 2019-12-02
Update google-analytics_analytics.js from https://github.com/gorhill/uBlock/commit/8a1a8b103f5

ryanbr:
Just sync recent patch from https://github.com/gorhill/uBlock/commit/8a1a8b103f5


AndriusA:
Looks good to me, but may not have the intended effect as the patched file is only used in tests.

The actual resources to be used by the browser are pulled in by the https://github.com/brave/brave-core-crx-packager  

Tagging @antonok-edm for the piece of documentation that explains this bit

PR #61, opened 2019-12-12
Add doc comments for cosmetic filter structs/methods

antonok-edm:
As per https://github.com/brave/brave-core/pull/3303#issuecomment-564898636 I wanted to make sure it's very clear how the cosmetic filtering interface is to be used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,8 +37,8 @@ base64 = "0.10"
 rmp-serde = "0.13.7"    # rmp-serde 0.14.0 breaks deserialization by changing how enums are deserialized
 hashbrown = { version = "0.6", features = ["serde"] }
 lifeguard = { version = "0.6", optional = true }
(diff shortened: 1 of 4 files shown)


antonok-edm:
Should be as simple as this:

```rust
let resources = engine.hostname_cosmetic_resources("github.com");

let stylesheet = engine.class_id_stylesheet(&["new_class1", "new_class2"], &["new_id1"], resources.exceptions.clone());
```

I'm noticing now that the signature for `Engine::hostname_cosmetic_resources` should just be updated to take `exceptions` by reference instead of by value. Once that's done, you'll even be able to remove the `.clone()` call.

```rust
let resources = engine.hostname_cosmetic_resources("github.com");

let stylesheet = engine.class_id_stylesheet(&["new_class1", "new_class2"], &["new_id1"], &resources.exceptions);
```

PR #63, opened 2020-01-09
Return list of selectors instead of a full stylesheet

antonok-edm:
Changes in `brave-core` to determine the first/third-partyness of affected page elements before applying a particular hide rule require access to the original selectors provided by the adblock engine. Building a stylesheet is now delegated to the caller.


antonok-edm:
Looks like CI failed because of `cargo audit`, I pushed #64 to address that

PR #64, opened 2020-01-10
Update reqwest to address cargo audit

antonok-edm:
The `http` crate recently had the following advisories released, as caught by `cargo audit`:
https://rustsec.org/advisories/RUSTSEC-2019-0033
https://rustsec.org/advisories/RUSTSEC-2019-0034

I've updated `reqwest` to `0.10`, which bumps `http` to a safely patched version.

### DIFF
--- a/benches/bench_redirect_performance.rs
+++ b/benches/bench_redirect_performance.rs
@@ -9,25 +9,48 @@ use adblock::request::Request;
 use adblock::blocker::{Blocker, BlockerOptions};
 use adblock::resources::resource_assembler::{assemble_web_accessible_resources, assemble_scriptlet_resources};
 
+use tokio::runtime::Runtime;
+
 lazy_static! {
     static ref PSL_LIST: psl::List = psl::List::new();
 }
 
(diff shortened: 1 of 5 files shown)


benches/bench_redirect_performance.rs
```diff
@@ -9,25 +9,48 @@ use adblock::request::Request;
...
+        .await
+        .iter()
         .flatten()
+        .cloned()
+        .collect()
+}
+
+/// Gets all rules with redirects, and modifies them to apply to resources at `a{0-n}.com/bad.js`
+fn get_redirect_rules() -> Vec<NetworkFilter> {
+    let mut async_runtime = Runtime::new().expect("Could not start Tokio runtime");
```
AndriusA: not really looked properly into async, is it necessary to depend on tokio as well? nothing in std?

antonok-edm: There needs to be some kind of executor, and the Rust standard library doesn't provide one. It's only a dev dependency here; we could try another one if needed but I haven't seen a big difference between them.

PR #68, opened 2020-02-04
Enable CSS validation for NPM package

antonok-edm:
The CRX packager will be generating serialized cosmetic rulesets; we want CSS validation there so we should enable it for the NPM package.


antonok-edm:
@AndriusA could you get this updated on NPM when you get a chance?

AndriusA:
you're ok with updating on NPM now, right?

antonok-edm:
Yes, should be all set

PR #72, opened 2020-02-06
Move to Fanboy Notification List

ryanbr:
Now the list is officially supported by Easylist, we can move to this list. Makes it easier to make changes for me without needing to do a PR's. Also seems Eyeo might be keen to [include it.](https://gitlab.com/eyeo/adblockplus/abpui/adblockplusui/issues/701)


ryanbr:
The difference as of 7th Feb between the Fanboy Notifications vs Brave Notifications is around 19 changes.

```
2 New Generic Blocks
2 New Generic Cosmetic
14 New Notification Servers 
1 New Whitelist.
```

PR #75, opened 2020-03-04
Include Polish Anti-adblock filters

ryanbr:
This includes the following polish filters from https://github.com/brave/ad-block/issues/146

Just need a check on language/title, used google to translate "Privacy" and "Supplemental".


ryanbr:
If you want to look over this @pes10k thanks :)

PR #79, opened 2020-03-24
add desc field for each filter list, remove duplicates and clean up existing regional filter list entries

pes10k:
- Makes the title field for each region formatted and capitalized more consistently
- Adds a description (desc) field for each region
- Remove duplicate Indonesian list
- Add / correct a few lang entires

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adblock"
-version = "0.2.4"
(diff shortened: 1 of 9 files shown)


emerick:
Do we need to bump the version number?

pes10k:
I do not know.  @bbondy or @AndriusA might know better?

bbondy:
yep but maybe @AndriusA can do that in a separate commit when he publishes. I defer to him.

PR #80, opened 2020-03-25
Add fuzzing using `cargo fuzz`, add test for resulting bug

ArniDagur:
Here I've added a fuzzing harness using `cargo fuzz`, along with a fuzzing target for `parse_filter`. Straight away, a bug was discovered, for which I've added a test case but not a fix: When the string `Ѥ` is input to the function, a crash occurs.

Ideally we would have a fuzzing target for each function in the crate's public interface, a lot of bugs can be found this way.

## Instructions
1. `cargo install cargo-fuzz`
2. `cargo fuzz run parse_filter`


ArniDagur:
@AndriusA Should I write more fuzzing targets?

AndriusA:
this is a great contribution and one I've been thinking we should do for some time. I just need to look through it properly and fix the bug you've uncovered. If you feel like doing more - great!

PR #67, opened 2020-01-25
Replace Prebake with Easylist Cookie

ryanbr:
Since Prebake Cookie hasn't been updated since 2018, we should replace this with Easylist Cookie.

Should help with cookie notices and resolve https://github.com/brave/brave-browser/issues/5318 Will be more affective when we roll out cosmetic filtering.

PR #82, opened 2020-03-27
Replace Prebake with Easylist Cookie

ryanbr:
Since Prebake Cookie hasn't been updated since 2018, we should replace this with Easylist Cookie.

Should help with cookie notices and resolve brave/brave-browser#5318 Will be more affective when we roll out cosmetic filtering.

Rebase of https://github.com/brave/adblock-rust/pull/67  (just had some merge issues)

Also re-did `component_id:` and `base64_public_key:`

PEM's also updated/uploaded.

### DIFF
--- a/src/filter_lists/regions.rs
+++ b/src/filter_lists/regions.rs
@@ -95,12 +95,12 @@ pub fn regions() -> Vec<FilterList> {
         FilterList {
             uuid: String::from("AC023D22-AE88-4060-A978-4FEEEC4221693"),
             url: String::from("https://easylist-downloads.adblockplus.org/easylist-cookie.txt"),
-            title: String::from("EU: Easylist-Cookie List - Filter Obtrusive Cookie Notices"),
+            title: String::from("Easylist-Cookie List - Filter Obtrusive Cookie Notices"),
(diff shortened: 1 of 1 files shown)


ryanbr:
If this looks okay @AndriusA  ?

ryanbr:
cc: @pes10k

pes10k:
Looks good to me.  Is the description still correct though for the new list?  Does EasyList Cookie also "Remove obtrusive EU cookie law notices"?

ryanbr:
It does. Either blocking the specific cookie javascript or hiding via cosmetic filtering.

pes10k:
Sounds good, just wasn't sure about the "EU" part

ryanbr:
Most 99% cookie mesages are EU related, but not exclusively (A few Russian, Australian and NZ sites) are covered.

pes10k:
Okie dokie, then could you slightly change the description for the list too?  Maybe just say something like "mostly targets EU etc"

ryanbr:
Probably easier to just remove the "EU" bit. so more of a generic "Removes obtrusive cookie law notices"

Issue #83, opened 2020-03-30
cosmetic filter parsing crash


AndriusA:
Unicode not always properly handled, crashing on `\##+js(,ݍ` (uncovered with fuzzing in https://github.com/brave/adblock-rust/pull/80)

Current WIP-branch for fuzz-testing uncovered issues is `fuzzing`

PR #84, opened 2020-03-30
Fuzzing

AndriusA:
Includes fuzzing contribution from @ArniDagur and fixes bugs uncovered with fuzzing

Closes https://github.com/brave/adblock-rust/issues/83


antonok-edm:
`parse_filter_failed_fuzz_4` is still failing because of a missing right paren. I'll push an update to that shortly

AndriusA:
not so much missing parent as insufficient care when jumping back by one byte in a string to check for things - there are multiple places in cosmetic filter parsing where those can cause issues, one I've fixed, left others

Issue #85, opened 2020-03-31
Questions about how to handle BlockerResult


ArniDagur:
I'm currently working on integrating Brave's adblocker in Qutebrowser. See https://github.com/qutebrowser/qutebrowser/pull/5317. I have a few questions regarding the `BlockerResult` struct:
```rust
pub struct BlockerResult {
    pub matched: bool,
    pub explicit_cancel: bool,
    pub important: bool,
    pub redirect: Option<String>,
    pub exception: Option<String>,
    pub filter: Option<String>,
    pub error: Option<String>,
}
```
* What do the string fields mean, and when are they not `None`?
* I guess that `redirect` is set to `Some(url)` when the browser should redirect the request somewhere else?
* If `matched` is `true`, does that mean that the browser should block the request?
* What are `explicit_cancel` and `important`?

Thank you very much for your help!

AndriusA:
Sorry, missed this.

> * What do the string fields mean, and when are they not `None`?

`redirect` is used as per [uBO's redirect syntax](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#redirect). If a rule has `redirect` option, this returns the body of the redirect to be injected.

`exception` is `Some` when the blocker matched on an exception rule. Effectively this means that there was a mach, but the request should not be blocked. It is a non-empty string if the blocker was initialized from list of rules with debugging enabled, otherwise the original string representation is discarded to reduce memory use.

`filter` similarly to exception, includes string representation of the rule when there is a match and debugging is enabled. Otherwise on a match contains `Some("")`

`error` only used to signal that there was an error in parsing the provided URLs when using the simpler `check_network_urls(url, frame_url, request_type)` method

> * I guess that `redirect` is set to `Some(url)` when the browser should redirect the request somewhere else?

explained above

> * If `matched` is `true`, does that mean that the browser should block the request?

We found that returning `200 OK` with an empty body generally causes less pages to break, so that's the default with `matched` is `true`, except if...

> * What are `explicit_cancel` and `important`?

... `explicit_cancel` is also true, in which case the request is cancelled.

`important` is used to signal that a rule with `important` option matched. `important` match means that exceptions should not apply and no further checking is necessary - the request should be blocked (empty body or cancelled). We keep separate instances of blocker for default lists and regional ones, so `important` here is used to maintain correct behaviour between them: checking should stop instead of moving to next instance iff `important` rule matched

ArniDagur:
[...]

PR #87, opened 2020-04-07
Add documentation for BlockerResult

ArniDagur:
I adapted @AndriusA's comment in #85 to document `BlockerResult`. Thank you very much!


ArniDagur:
Thanks @antonok-edm! Could you perhaps also look at https://github.com/brave/adblock-rust/issues/88 while you're at it?

PR #89, opened 2020-04-13
Replace ABP-Japanese with Adguard Japanese (brave/adblock-lists#355)

ryanbr:
Based on: https://github.com/brave/adblock-lists/issues/355 request. The most comprehensive list and also used by uBO as the default Japanese filter. 

Since ABP Japanese is not well supported and has many problematic filters; This change will help fix a few work arounds we needed to make (which will need to be reverted) 

https://github.com/brave/adblock-lists/pull/366
https://github.com/brave/adblock-lists/pull/319
https://github.com/brave/adblock-lists/pull/274
https://github.com/brave/adblock-lists/pull/270

More Japanese filter issues, https://github.com/brave/adblock-lists/pulls?q=is%3Apr+is%3Aclosed+label%3Aregional

### DIFF
--- a/src/filter_lists/regions.rs
+++ b/src/filter_lists/regions.rs
@@ -214,7 +214,7 @@ pub fn regions() -> Vec<FilterList> {
         },
         FilterList {
             uuid: String::from("843B0AB7-073C-4007-8747-A4CD88B8124D"),
-            url: String::from("https://filters.adtidy.org/extension/ublock/filters/7.txt"),
+            url: String::from("https://filters.adtidy.org/ios/filters/7.txt"),
             title: String::from("Adguard Japanese filters (日本用フィルタ)"),
             langs: [String::from("ja")].to_vec(),
             support_url: String::from("https://github.com/AdguardTeam/AdguardFilters"),
--- a/src/filter_lists/regions.rs
+++ b/src/filter_lists/regions.rs
@@ -223,16 +223,6 @@ pub fn regions() -> Vec<FilterList> {
             desc: String::from("Removes advertisements from Japanese websites")
         },
         FilterList {
-            uuid: String::from("1E6CF01B-AFC4-47D2-AE59-3E32A1ED094F"),
-            url: String::from("https://raw.githubusercontent.com/gfmaster/adblock-korea-contrib/master/filter.txt"),
-            title: String::from("Korean Adblock List"),
-            langs: [String::from("ko")].to_vec(),
-            support_url: String::from("https://github.com/gfmaster/adblock-korea-contrib"),
-            component_id: String::from("jboldinnegecjonfmaahihagfahjceoj"),
-            base64_public_key: String::from("MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEArcE90V2GoQ9InvoiVqxsEnhHhdJz68yyWA+HzQBxcxitqfF5Vu6gRiKB3+W/iTJ9XNE449krWrYAjDqVt6GF8ilpieIngwCNsl7jP5RRhrcrk65as4bawSimodit+TVi7ZpIFDWoalj5RIO+rjKfwkkBHujeM7qYr/Vm1NXre57ea80TvbdoA4XgnAYHKY5VV/WstFL8FR3Do+38EyaIzKfoiRrV97BoC440oifNdi0nJQhIgULfNjUolOh9eAQKuymId+WjZeSKSckUyKMQsQ0VVDjCbm5mwRqC5D+MT9L/8sKcdXBXuGrXyaZzp3eOmc41q01VkRLXCTrx37hGQwIDAQAB"),
-            desc: String::from("Removes advertisements from Korean websites")
-        },
-        FilterList {
             uuid: String::from("45B3ED40-C607-454F-A623-195FDD084637"),
             url: String::from("https://raw.githubusercontent.com/yous/YousList/master/youslist.txt"),
             title: String::from("YousList"),
--- a/src/filter_lists/regions.rs
+++ b/src/filter_lists/regions.rs
@@ -213,13 +213,13 @@ pub fn regions() -> Vec<FilterList> {
             desc: String::from("Removes additional advertisements from Italian websites, may break some websites")
         },
         FilterList {
-            uuid: String::from("03F91310-9244-40FA-BCF6-DA31B832F34D"),
-            url: String::from("https://raw.githubusercontent.com/k2jp/abp-japanese-filters/master/abpjf.txt"),
-            title: String::from("ABP Japanese filters (日本用フィルタ)"),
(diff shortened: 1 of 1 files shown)


ryanbr:
* PEM was also uploaded also.

* Filter has some Adguard specific filters.
`!+ PLATFORM(ext_ff, ext_opera, ios, ext_android_cb, ext_ublock)`
`!+ NOT_PLATFORM(ext_safari)`

Doesn't look too major, but does these platform checks work for us? @antonok-edm   cc: @pes10k

pes10k:
@ryanbr I'm all but certain that we don't we have any support for those sorts of filters (e.g. PLATFORM / NOT_PLATFORM)

Alex-302:
`PLATFORM` directive used during filters comtilation only, not parsed by ad blocker. 
For example AdGuard Japanese filter for uBO has only supported rules(directive+rule)
https://filters.adtidy.org/extension/ublock/filters/6.txt 

`!+ NOT_PLATFORM(windows, mac, android)` means that the rule will be available for all platforms, except AdGuard **APPS** for Windows/MacOS/Android.

DandelionSprout:
`(日本用フィルタ)` is not normally a part of *AdGuard Japanese Filters*' official title, so I looked into AdGuard's CrowdIn project to see if it was part of the Japanese translation of the name, which it was: 
image

ryanbr:
Translated: `日本用フィルタ` is just `Filter for Japan` Since it should still be applicable to the new list.

pes10k:
@antonok-edm how much work would it be to add in support for the additional formats used in this list? https://filters.adtidy.org/extension/chromium/filters/7.txt

Specifically, the #%# and `PLATFORM / NOT_ PLATFORM` rules? 

@ryanbr how useful would this list be if (short to medium term) we excluded rules that used the above instructions?

ryanbr:
We could just ignore `#$#` @antonok-edm .  The `! PLATFORM` checks are in a comment, so it'll be safely ignored by us

antonok-edm:
`#%#` would actually be pretty easy to represent as just a templated scriptlet where the single argument is the entire scriptlet. But it seems dangerous in general as it would allow arbitrary JS injection that completely bypasses the reviews we're doing on uBlock's scriptlets.

`PLATFORM` / `NOT_PLATFORM` would be tougher to implement as we'd need to either write and serve platform-specific DAT files (seems to be how AdGuard handles it), or patch every API in `adblock-rust` with an argument to specify the platform it's running on.

pes10k:
@antonok-edm i think all that'd be needed is to ignore most of them, maybe with the exception of `ext_ublock`, which we'd just allow.  No need to do any platform matching stuff, could all be done at parse/serialization time.

Alex-302:
If supported basic ABP syntax only, you can use https://filters.adtidy.org/ios/filters/7.txt - there are no JS and CSS rules.

Alex-302:
@antonok-edm
> seems to be how AdGuard handles it

AdGuard hangles them during filters compilation, not in ad blocker(compare that file with filters source and you will see the difference)

@ryanbr Is not CSS rules(uBO syntax) supported?

ryanbr:
@Alex-302  Yeah, `##+js(..)`  style rules are supported in upcoming Brave changes (not yet in Brave release).  Just `#$#` would need changes on our end. But as an interim fix we could just use the ABP Syntax, seems a pretty decent list.

ryanbr:
@antonok-edm is it easier to use the IOS list (https://filters.adtidy.org/ios/filters/7.txt), or ignore the specific Adguard (`#$#`, `#%#`) filters in Brave?

pes10k:
Another option would be to treat this list the way we treat uBO resources, and pull it in occasionally and audit it, if the `#%#` etc rules are important

ryanbr:
Since they're combing a few lists, we could just mirror our own copy of https://filters.adtidy.org/extension/ublock/filters/6.txt and just comment out unsupported rules. (and create PR's to sync)

pes10k:
I think that could be good, though will make it more difficult to keep things in sync.  Lets do this (might require upstream changes):

1. if @antonok-edm thinks its not a big change, add support for the additional rule formats
2. split this out into two lists, one thats just blocking rules, and another that includes JS injection (which would require security reviews)
3. pull the blocking rules automatically
4. pull the JS injecting rules as needed, on review

Would that work?  If so, is it something the adguard folks could do upstream (if its not already possible).  Maybe they'd be okay with the slight additional work, if it meant more users for this list (and so, potentially more support / crowd-sourcing).   WDYT?

ryanbr:
Looks like we're ignore various Adguard filters `#%#,#$#,#?#..`. https://github.com/brave/adblock-rust/blob/master/src/lists.rs#L156 and we'll be ignoring `! Platform..` https://github.com/brave/adblock-rust/blob/master/src/lists.rs#L135  It might be safe to deploy the /ublock/ filter.  Am I correct to confirm these @antonok-edm ?

antonok-edm:
@ryanbr Yep, that's correct - anything under `! PLATFORM...` will be parsed and included, and the `#%#,#$#,#?#...` filters would just be ignored.

This looks like the first list we're including with Unicode characters, so it'll be important that the DAT generator is using the most recent (v0.2.7) adblock-rs.

pes10k:
@ryanbr okie, so after reviewing this with @antonok-edm, i think that implementing most of adguard syntax would be more trouble than its worth at the moment.  (there are also some issues with injecting global variables into the page that seem worrisome in the AG approach that we'd need to address if we were going to start pulling in their lists.)

So, the short term options then are:
1. Keep the current JP list
2. Switch to the adguard iOS list, which (with the exception of PLATFORM) doesn't have AG specific syntax
3. Move to some third / other list

Of those options, which do you think would be the best route forward @ryanbr ?

ryanbr:
@pes the Adguard iOS list should be safer bet, the current ABP Japanese list is causing too many false positives to keep. (So I prefer 2.) The `! Platform` bits will be ignored, and seems to be applying cosmetics below it (which we support).

Will fix (remove) brave-unbreak abp-japanese filters once this lands into Brave Release. Happy to land this week if @pes10k @antonok-edm are okay with it

pes10k:
That sounds fine to me.  One last Q from above though is whether this should be a new list, that JP users would need to select (so, need to change PEM), or should existing JP list subscribers / users just automatically start using this new list (so, no new PEM)?

My vote is for 2, but I dont have a good understanding of how well this list substitutes for the other

ryanbr:
Sure, we can re-use the same PEM. I just changed it the PEM out of habit when changing or adding new lists. Could push to get this into Brave Release also

pes10k:
If we don't update the PEM, i think the change will automatically be pulled into brave release, no?  I'll also double check internally, but I believe thats the case

ryanbr:
Also attached removal of Outdated Korean list, https://github.com/gfmaster/adblock-korea-contrib/commits/master hasn't been updated for a while.  

uBO is using YousList as default and we should do also, No other changes needed, we have YousList  already in our regional list.

PR #92, opened 2020-04-26
Ignore new :upward prefix

ryanbr:
Recent commits in https://github.com/gorhill/uBlock/wiki/Procedural-cosmetic-filters#subjectupwardarg and https://github.com/gorhill/uBlock/commit/72bb70056843024b1a31fe1ab9c90bd4e8260ba2

This is replacing 'nth-ancestor`, but its probably safe to keep both in the ignore list.


ryanbr:
Since it's being rolled out now, we should ignore them until its supported.

Does this commit look good? @antonok-edm

PR #96, opened 2020-05-17
Updated NoCoin Filter List

TheAresjej:
2018 vs 2020

### DIFF
--- a/data/brave/coin-miners.txt
+++ b/data/brave/coin-miners.txt
@@ -1,9 +1,9 @@
 [Adblock Plus 2.0]
 !-------------------------------------------------------------------------------
 ! Title: NoCoin Filter List
-! Expires: 2 days (update frequency)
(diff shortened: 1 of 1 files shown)


ryanbr:
I'll review commit this week, Looking at merging fixes from https://github.com/hoshsadiq/adblock-nocoin-list possibly instead.

TheAresjej:
@ryanbr what do you think of it? 😄

ryanbr:
How was the diff created @TheAresjej  ?

TheAresjej:
> How was the diff created @TheAresjej ?

I manually compared the changes and added new entries but in which case, if this update is not useful, it can be closed 😄

ryanbr:
I'll re-checkout vs the current nocoin version, thanks for the reminder for requesting an update to the coin list :)

Issue #1, opened 2019-03-19
unhandled options


AndriusA:
A few options are not currently handled from easylist and easyprivacy:

- [ ] `popup` - Include pages opened in a new tab or window
- [x] `generichide` - Used to prevent applying global element rules on a page (e.g. `@@||example.com^$generichide`)
- [x] `document` - Used to whitelist the page itself (e.g. `@@||example.com^$document`)
- [ ] `webrtc` - Include or exclude connections opened via RTCPeerConnection instances to ICE servers
- [ ] `elemhide` - Used to prevent element rules from applying on a page (e.g. `@@||example.com^$elemhide`)
- [ ] `genericblock` - Used to prevent applying global blocking rules on a page (e.g. `@@||example.com^$genericblock`)

AndriusA:
For network filters, there are still a few rule options that are not supported, wrt https://github.com/brave/ad-block/issues/217

- `popunder`, `popup` - ad-blocker does not directly interact with popup/popunder blocking
- `csp` - rules are parsed but `csp` checking and handling on the browser side is not implemented
- `inline-script` - inline scripts are normally blocked through CSP which is not currently handled
- `generichide` - `generichide` is a cosmetic filtering-related option, and as such using it has no negative consequence with regard to privacy since cosmetic filtering has no privacy value.
- `match-case` - all rules are considered case insensitive

antonok-edm:
See also - https://github.com/brave/brave-browser/issues/8478

Emi-TheDhamphirInLoveUnderTheFrozenStar:
I noticed uBlock have added some new features like `$to` or a new alias for $domain: `$from`, and expanded regex support, so I made this quick table to compare uBlock features (from their wiki) Brave doesn't partially or completely support.
Most of these incompatibilities are already mentioned here and other issues, but I also added workarounds I use to fix the limitations on my side. Just a reminder of how Brave is doing against uBlock in 2023, 4 years after this issue was opened.

Some of these features might not seem too relevant or used even today, but since Brave uses uBlock lists as default they are important to support, because when you put all the unsupported rules together, they represent an important amount of rules Brave is currently ignoring, which obviously reduce the potential of Brave to reduce tracking and avoid bad popups and all that in many websites.


|     | Feature                                                                                                                                                                                  | If Partial, What's currently not supported                                                                                                                                                      | Examples                                                                                                                                                                                        | Workarounds                                                                                                                                                                                                                                                                                                                                                      |
[...]

PR #97, opened 2020-05-18
Support generichide rules

antonok-edm:
Adds support for `$generichide` exception rules (as well as the corresponding uBO-style alias, `$ghide`). These are network rules that prevent generic cosmetic element hiding from being applied to matching URLs.

This partially addresses https://github.com/brave/adblock-rust/issues/1.

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -98,6 +98,9 @@ pub struct Blocker {
     #[cfg(feature = "object-pooling")]
     #[serde(skip_serializing, skip_deserializing)]
     pool: TokenPool,
+
+    #[serde(default)]
+    generic_hide: NetworkFilterList,
 }
 
 impl Blocker {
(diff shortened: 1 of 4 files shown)


src/blocker.rs
```diff
@@ -1346,14 +1389,13 @@ mod legacy_rule_parsing_tests {
     // easyList = { 24478, 31144, 0, 5589 };
     // not handling (and not including) filters with the following options: 
     // - $popup
-    // - $generichide
     // - $subdocument
```
AndriusA, line 1392: not about this PR, but I think we actually do handle `$subdocument`

antonok-edm: looks like `$document` right below is handled too. I'll go ahead and remove both.

AndriusA: `$document` only somewhat - there is a spec for handling main document url blocking, but not implemented

antonok-edm: ah oops, I'll add that back then :sweat_smile:

PR #130, opened 2020-08-24
Replace Rc with Arc

ArniDagur:
This allows me to build [`python-adblock`](https://github.com/ArniDagur/python-adblock) with `PyO3` version `0.11.1` without any thread safety errors, if I disable the `object-pooling` feature.

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -7,8 +7,7 @@ use std::fmt;
 use crate::request;
 use crate::utils;
 use crate::utils::Hash;
-use std::cell::RefCell;
-use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -4,2125 +4,2123 @@
 name = "adblock"
 version = "0.3.1"
 dependencies = [
- "addr 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
```
antonok-edm, line 7: Not sure what you did to the `Cargo.lock` file - would you mind reverting the changes here?

ArniDagur: I'm not sure either! I've reverted the changes.

PR #134, opened 2020-09-01
Fix filters not work with ports number.

mOnkD404:
The filter does not work properly when processing rules like "||abc.com:8080". The parsed filter will not match url "https://www.abc.com:8080/". Because the parsed filter has host name "abc.com:8080" , and the parsed url request has  host name "abc.com" without port number. I tried to fix it , and also wrote some test cases.

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -463,41 +463,15 @@ impl NetworkFilter {
                 let slash_index = twoway::find_str(&line[filter_index_start..], "/");
                 slash_index
                     .map(|i| {
-                        // Handle filters with port number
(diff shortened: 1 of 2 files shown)


antonok-edm (changes requested):
Thanks for this! I definitely would like `adblock-rust` to handle ports correctly.

It looks like your solution here is to add the port to the "path" part of the URL - e.g., `||test.com:8080/test` gets interpreted as having `hostname: Some("test.com")` and `filter: Simple(":8080/test")`. This is pretty inconsistent with the URL spec, and I believe it will cause issues with detecting third-partyness.

Could you take another look at this and see if it's possible to include the port with the hostname instead?

PR #157, opened 2021-03-03
update addr from 0.2 to 0.9

rushmorem:


src/url_parser/mod.rs
```diff
@@ -29,12 +32,10 @@ impl ResolvesDomain for DefaultResolver {
         if host.is_empty() {
             (0, 0)
         } else {
-            match host.parse::<addr::DomainName>() {
+            match Name::try_from(host) {
```
rushmorem, line 35: For performance reasons, `addr` no longer depends on the `idna` crate so it no longer supports IDNs out of the box. I'm not familiar with the internals of this library so I'm not sure if it is something I need to handle here. I saw `idna::domain_to_ascii` being called somewhere in the library so I'm hoping that the input to `ResolvesDomain::get_host_domain` is an ASCII encoded string. Kindly correct me if I'm wrong.

src/url_parser/mod.rs
```diff
@@ -29,14 +31,12 @@ impl ResolvesDomain for DefaultResolver {
         if host.is_empty() {
             (0, 0)
         } else {
-            match host.parse::<addr::DomainName>() {
+            match domain::Name::parse(host) {
```
rushmorem, line 34: For performance reasons, `addr` no longer depends on the `idna` crate so it no longer supports unicode domains out of the box. I'm not familiar with the internals of this library so I'm not sure if it is something I need to handle here. I saw `idna::domain_to_ascii` being called somewhere in the library so I'm hoping that the input to `ResolvesDomain::get_host_domain` is an ASCII encoded string from that function. Kindly correct me if I'm wrong.

antonok-edm:
Excellent, thank you for your work on this @rushmorem!

rushmorem:
My pleasure :)

Issue #169, opened 2021-04-30
Support uBO redirect rules


ryanbr:
From: https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#redirect

1. A filter with redirect blocks and redirects the mentioned request.
2. A filter with redirect-rule only redirects the mentioned request if it is blocked by any other filter. 
"redirect is internally translated to redirect-rule"

**Testcase:** `https://turreta.com/2020/11/14/rust-array-size-limit-to-32-no-more-in-1-47-0-release/`
**Affected filter:** `||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$script,redirect-rule=googlesyndication_adsbygoogle.js`

Currently only working around with `@@||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$script,domain=turreta.com`

376 redirect-rule filters in uAssets.

We have a few workarounds in unbreak: https://github.com/brave/adblock-lists/blob/master/brave-unbreak.txt#L256

antonok-edm:
Keeping track of implementation here:
- [ ] DAT packager side, begin duplicating any `redirect` filter, and adding it to `Blocker::redirects`. This should not change any behavior for existing clients.
- [ ] When updated DATs are available, begin treating redirects as independent from blocking/exception filters. This means exceptions for blocking should only be looked for in `Blocker::exceptions`, and exceptions for redirections should only be looked for in `Blocker::redirects`. `Blocker::redirects` should be checked for _all_ matches rather than the first match (similar to `$csp`), and results should be combined accordingly. At this point, a redirection exception should always occur whenever a blocking exception occurs, but the code should be ready to handle the case when they do not occur together.
- [ ] Once clients have updated sufficiently, support for `redirect-rule` can be added on the DAT packager side. These will be added only to `Blocker::redirects`. At the same time, support can be added for [`redirect-rule=resource:[priority]` syntax](https://github.com/gorhill/ublock/wiki/Static-filter-syntax#redirect).

PR #184, opened 2021-09-02
`redirect-url` filter option

ShivanKaul:
Add a new filter option to `adblock` called `redirect-url` (similar to `redirect=`) that allows loading of replacement resources via a URL instead of having to bundle all replacement resources in the client.

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -940,7 +940,7 @@ mod tests {
             let filters = vec!["||foo.com"];
             let network_filters: Vec<_> = filters
                 .into_iter()
-                .map(|f| NetworkFilter::parse(&f, true))
+                .map(|f| NetworkFilter::parse(&f, true, Default::default()))
                 .filter_map(Result::ok)
                 .collect();
             let filter_list = NetworkFilterList::new(network_filters, false);
@@ -952,7 +952,7 @@ mod tests {
             let filters = vec!["||foo.com", "||bar.com/foo"];
             let network_filters: Vec<_> = filters
                 .into_iter()
-                .map(|f| NetworkFilter::parse(&f, true))
+                .map(|f| NetworkFilter::parse(&f, true, Default::default()))
                 .filter_map(Result::ok)
                 .collect();
             let filter_list = NetworkFilterList::new(network_filters, false);
@@ -970,7 +970,7 @@ mod tests {
             let filters = vec!["||foo.com", "||foo.com/bar", "||www"];
             let network_filters: Vec<_> = filters
                 .into_iter()
-                .map(|f| NetworkFilter::parse(&f, true))
+                .map(|f| NetworkFilter::parse(&f, true, Default::default()))
                 .filter_map(Result::ok)
                 .collect();
             let filter_list = NetworkFilterList::new(network_filters, false);
@@ -989,7 +989,7 @@ mod tests {
             let filters = vec!["||foo.com", "||foo.com$domain=bar.com"];
             let network_filters: Vec<_> = filters
                 .into_iter()
-                .map(|f| NetworkFilter::parse(&f, true))
+                .map(|f| NetworkFilter::parse(&f, true, Default::default()))
                 .filter_map(Result::ok)
                 .collect();
             let filter_list = NetworkFilterList::new(network_filters, false);
@@ -1012,7 +1012,7 @@ mod tests {
             let filters = vec!["foo*$domain=bar.com|baz.com"];
             let network_filters: Vec<_> = filters
                 .into_iter()
-                .map(|f| NetworkFilter::parse(&f, true))
+                .map(|f| NetworkFilter::parse(&f, true, Default::default()))
                 .filter_map(Result::ok)
                 .collect();
             let filter_list = NetworkFilterList::new(network_filters, false);
@@ -1049,7 +1049,7 @@ mod tests {
     fn test_requests_filters(filters: &Vec<&str>, requests: &Vec<(Request, bool)>) {
         let network_filters: Vec<_> = filters
             .into_iter()
-            .map(|f| NetworkFilter::parse(&f, true))
+            .map(|f| NetworkFilter::parse(&f, true, Default::default()))
             .filter_map(Result::ok)
             .collect();
         let filter_list = NetworkFilterList::new(network_filters, false);
@@ -1214,7 +1214,7 @@ mod tests {
 mod blocker_tests {
 
     use super::*;
-    use crate::lists::{parse_filters, FilterFormat};
+    use crate::lists::{parse_filters, FilterFormat, parse_filters_with_opts, ParseOptions};
     use crate::request::Request;
     use std::collections::HashSet;
     use std::iter::FromIterator;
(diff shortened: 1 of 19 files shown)


src/blocker.rs
```diff
@@ -234,12 +243,31 @@ impl Blocker {
...
+            // Filter redirect option is set
+            // Filter redirect option is set
+            if let Some(redirect) = f.redirect_url.as_ref() {
+                // Should be a valid URL
+                let maybe_parsed_url = parse_url(&redirect);
+                if maybe_parsed_url.is_some() {
+                    Some(maybe_parsed_url.unwrap().url)
+                } else {
+                    // Invalid URL!
+                    eprintln!("Matched rule with redirect-url option but URL not formatted correctly");
```
ShivanKaul, line 256: @antonok-edm do you think this should error out if the passed in value for `redirect-url=` is not parsed as a URL? We don't seem to do that elsewhere in this function (check TODO on line 236)

antonok-edm: could you move this check to occur at list parsing time? The more we can avoid parsing things during matching the better.

antonok-edm (changes requested):

src/blocker.rs
ShivanKaul: Actually nvm, I'm going to move this parse logic to network so it errors out on filter parse time, not blocker match time.

ShivanKaul: Updated!

src/filters/network.rs
```diff
@@ -381,6 +398,7 @@ pub struct NetworkFilter {
     pub opt_domains: Option<Vec<Hash>>,
     pub opt_not_domains: Option<Vec<Hash>>,
     pub redirect: Option<String>,
+    pub redirect_url: Option<String>,
```
antonok-edm, line 401: Sadly we can't just add fields in the middle here; this struct definition is directly used in the `v0` format so we'd have to do the same thing with `#[serde(default)]` and add it as the last field of the struct.

Luckily though, since `redirect` and `redirect_url` are mutually exclusive, we do have another option. We can use the original `redirect` field in both cases, and use an unused bit in `NetworkFilterMask` to signify the redirect semantics. So far bits 15, 26, and 31 are unused - you can go ahead and repurpose bit 15 (there used to be a `$fuzzy` option, but the fuzzy matching logic was never implemented). This approach has the added benefit of not increasing the in-memory size of `NetworkFilter`, which is a valid concern when loading thousands of rules.

antonok-edm (commented):
Custom filter list subscriptions is merged in `brave-core` now, so there's no way this can make it to the browser unless there's some way to ensure `redirect-url` rule support can be disabled when parsing a list. I'd recommend adding some kind of `ParseOptions` struct in `src/lists.rs`, which can be passed as an argument to a new method on `FilterSet`, something like `add_filter_list_with_opts` and a similar `add_filters_with_opts`. The existing `FilterFormat` arg can be part of the options struct, as well as a new option to disable any `redirect-url` rules.

One last consideration that doesn't require any immediate action - I am working in parallel on `redirect-rule` support, which can provide a redirection for a matching request without causing it to be blocked (i.e. it will only be blocked+redirected if it _also_ matches a separate blocking rule). For now it looks like `redirect-url` will have to have the same behavior as `redirect`, implementation-wise, but we should move forward with list-building as if it has the behavior of `redirect-rule`. `redirect` is kind of the old way of doing it; the separation of concerns is better with `redirect-rule` semantics and it also keeps us from having to support something like `redirect-url-rule`.

Edit: I've posted my `redirect-rule` implementation plan/progress [here](https://github.com/brave/adblock-rust/issues/169#issuecomment-916471765).

src/blocker.rs
```diff
@@ -344,7 +366,11 @@ impl Blocker {
 
                 // Redirects are independent of blocking behavior.
                 if filter.is_redirect() {
-                    redirects.push(filter.clone());
+                    let mut filter = filter.clone();
+                    if filter.is_redirect_url() {
+                        filter.mask.set(crate::filters::network::NetworkFilterMask::IS_REDIRECT_URL, true);
```
antonok-edm, line 371: Updating the mask should occur at parse-time, not when adding to the engine.

<sub>(I know, I just finally removed a mask update here regarding `important` - there was no way around that when initially implemented... But this should work fine at parse time.)</sub>

src/data_format/legacy.rs
```diff
@@ -284,6 +284,7 @@ struct DeserializeFormatRest {
 
     #[serde(default)]
     generic_hide: NetworkFilterListLegacyDeserializeFmt,
+
```
antonok-edm, line 287: nit: extra line

src/data_format/legacy.rs
```diff
@@ -230,7 +230,7 @@ impl DeserializeFormat {
         let rest = match rmps::decode::from_read(&mut gz) {
             Ok(rest) => rest,
             Err(ref e) if is_eof_error(e) => Default::default(),
-            Err(e) => return Err(e.into()),
+            Err(e) => return Err(e.into())
```
antonok-edm, line 233: nit: removed comma here

src/data_format/v0.rs
```diff
@@ -25,6 +25,7 @@ pub(crate) struct SerializeFormat<'a> {
     exceptions: &'a NetworkFilterList,
     importants: &'a NetworkFilterList,
     redirects: &'a NetworkFilterList,
+
```
antonok-edm, line 28: nit: extra line

src/filters/network.rs
```diff
@@ -248,7 +253,8 @@ struct AbstractNetworkFilter {
 }
 
 impl AbstractNetworkFilter {
-    fn parse(line: &str) -> Result<Self, NetworkFilterError> {
+    fn parse(line: &str, opts: ParseOptions) -> Result<Self, NetworkFilterError> {
+
```
antonok-edm, line 257: nit: extra line

src/lib.rs
```diff
@@ -1,6 +1,8 @@
 #![allow(dead_code)]
 #![forbid(unsafe_code)]
 
+extern crate url;
+
```
antonok-edm, line 5: `extern crate` is not required for Rust 2018 edition and beyond

src/lists.rs
```diff
@@ -297,17 +326,28 @@ pub fn parse_filter(
...
-    list: &[String],
+
+/// Parse a single filter rule
+pub fn parse_filter(
+    line: &str,
     debug: bool,
     format: FilterFormat,
+) -> Result<ParsedFilter, FilterParseError> {
+    return parse_filter_with_opts(line, debug, ParseOptions {format, ..Default::default()});
+
```
antonok-edm, line 337: nit: extra line

src/lists.rs
```diff
@@ -317,6 +357,16 @@ pub fn parse_filters(
     (network_filters, cosmetic_filters)
 }
 
+/// Parse an entire list of filters, ignoring any errors
+pub fn parse_filters(
+    list: &[String],
+    debug: bool,
+    format: FilterFormat,
+) -> (Vec<NetworkFilter>, Vec<CosmeticFilter>) {
+
```
antonok-edm, line 366: nit: extra line

@ShivanKaul how'd you manage to insert so many of these? :sweat_smile:

antonok-edm (commented):
Ok, we are _almost_ there.

I can take it from here; I'm going to rebase this on top of https://github.com/brave/adblock-rust/pull/186 and add a few minor tweaks to make the API a little less crufty (i.e. consolidating the buildup of existing constructors and options where possible) since it looks like we're going to have to make a minor version bump with breaking changes anyways.

src/lists.rs
ShivanKaul: Haha sorry - I'm going to blame my editor. We should use a standard format config that we check in.

src/blocker.rs
ShivanKaul: I think this is extraneous actually, I already update the mask in network.rs

ShivanKaul:
@antonok-edm addressed nits and removed mask setting in `blocker.rs`

ShivanKaul:
I asked uBO folks what their thoughts were on `redirect-url`, seems like they understandably don't want to implement because of security concerns: https://www.reddit.com/r/uBlockOrigin/comments/poy9us/interest_in_redirecturl_filter_option/
@antonok-edm for us, we'd only be enabling the redirect-url option (via ParseOptions) when reading the sugarcoat filterlist, correct?

antonok-edm:
> don't want to implement because of security concerns

Yep, sounds about right. I imagine it'd also be harder for them to use such a thing without any server infrastructure.

> we'd only be enabling the redirect-url option (via ParseOptions) when reading the sugarcoat filterlist, correct?

Correct, that's exactly why I wanted the option added!

uBlock-user:
@antonok-edm No, because of security/performance issues(amoung others issue/s) raised here - https://github.com/uBlockOrigin/uBlock-issues/issues/46#issuecomment-391303700

ABP also implemented it and once these security issues came to light in the public, they immediately moved forward to remove it.

uBlock-user:
@antonok-edm No, because of security/performance issues(amoung others issue/s) raised here - https://github.com/uBlockOrigin/uBlock-issues/issues/46#issuecomment-391303700

ABP also implemented it and once these issues came to light in the public, they immediately moved forward to remove it.

antonok-edm:
@uBlock-user I appreciate the context. Our `$redirect-url` implementation is a little bit different from `$rewrite` though - it's a redirect to a single fixed URL (i.e. no regexes or other runtime computations/modifications), and at least for now we're only allowing this to be used in Brave lists (any rules using it will get filtered out from 3rd party lists).

We plan on using this for very similar use-cases to `redirect`, except that the resources are [autogenerated privacy-preserving alternatives to scripts observed on real websites](https://www.peteresnyder.com/static/papers/sugarcoat-ccs-2021.pdf). This should be a solid improvement in cases where bad scripts need exceptions to prevent sites from breaking. In practice we've already used this approach to stub out ~100MB of scripts, and they may need to be updated occasionally to keep up with website changes; it doesn't make sense to ship _all_ of these resources to every user.

antonok-edm (approved):

PR #185, opened 2021-09-09
Always duplicate redirect rules into `Blocker::redirects`

antonok-edm:
First implementation step of [this plan](https://github.com/brave/adblock-rust/issues/169#issuecomment-916471765).

> DAT packager side, begin duplicating any redirect filter, and adding it to `Blocker::redirects`. This should not change any behavior for existing clients.


antonok-edm:
@ShivanKaul heads up. I don't think this change will significantly impact #184, but part 2 will take some special attention. I'll work on the part 2 PR in parallel and I can help you update your code if necessary.

PR #191, opened 2021-10-08
Restrict `redirect-url` to only load over HTTPS

ShivanKaul:
Given that we're loading arbitrary resources over the network, even if the filter lists are controlled by us, we should only load over HTTPS to prevent MITM attacks.


antonok-edm:
Closing this as of #235

PR #192, opened 2021-10-22
Update to 2021 edition

antonok-edm:
No code changes required :tada: 

Will wait on merging at least until Brave's CI pipelines are updated to use Rust 1.56.


antonok-edm:
Closing in favor of #300

Issue #203, opened 2022-03-28
Optimize regexp memory usage


atuchin-m:
From https://github.com/brave/brave-browser/issues/21603#issuecomment-1076289536

Adblock filters take +40Mb (Private Memory Footprint) even without extra filters.
Most of that memory is consumed by heap allocation with the stack https://github.com/brave/brave-browser/issues/21603#issuecomment-1076289536

35MB is taken by HashMaps that store `NetworkFilter`. Each `NetworkFilter` takes ~400 bytes which looks much more than the text version.
The most expensive field is `regexp`.

PR #204, opened 2022-03-28
Optimize regexp memory usage

atuchin-m:
Resolves https://github.com/brave/adblock-rust/issues/203
Currently, the regexp field consumes a lot of memory because of the size of `Arc<RwLock<Option<Arc<CompiledRegex>>>>`.
It takes 40 bytes per item even when None is stored.

Most of the rules is non-regex(~90%).
This new approach takes 8 bytes per item for such rules.
It saves about 7Mb of memory in total.

Important note: this makes the engine not thread-safe and wrap under a new feature.
In fact, the browser uses adblock from a single sequence => we don't need to synchronize threads, the `thread-safety` feature should be disabled in browser to get benefit from the optimization.

```
Memory usage after loading data/rs-ABPFilterParserData.dat in tests/deserialization.rs
(Release, ubuntu x64, added jemallocator locally to calculate):
before: 42094712 bytes allocated/53563392 bytes resident after: 34332128 bytes allocated/48881664 bytes resident
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling"]
+default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "thread-safety"]
 full-domain-matching = []
 metrics = []
 full-regex-handling = []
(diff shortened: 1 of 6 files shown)


atuchin-m:
@antonok-edm I'm not sure about enabling the feature in `native/Cargo.toml`.
Looks like it's used only in node-js and multi-threading doesn't need here.

antonok-edm (approved):
I pushed a few minor fixes, including:
- 4 space indentation as per standard Rust formatting
- removed manual `default()` implementation on `RegexStorage`
- related the feature to `object-pooling` as an opt-out optimization
- added static assertions for `Send` and `Sync` implementations on `Engine`

This should be all set now, I will merge it and create a new release

antonok-edm:
Published as [v0.5.2](https://crates.io/crates/adblock/0.5.2) on crates.io

PR #211, opened 2022-06-02
Add BraveServiceKey to live DAT tests

antonok-edm:
The endpoint for downloading the most recent DAT file was recently updated to require a service key; this allows the tests which required a live DAT to continue working by setting the `$brave_service_key` environment variable.

### DIFF
--- a/tests/live.rs
+++ b/tests/live.rs
@@ -106,10 +106,19 @@ fn get_blocker_engine_deserialized() -> Engine {
     use futures::FutureExt;
     let async_runtime = Runtime::new().expect("Could not start Tokio runtime");
 
+    let brave_service_key = std::env::var("brave_service_key")
+        .expect("Must set the $brave_service_key environment variable to execute live tests.");
+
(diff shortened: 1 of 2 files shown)


tests/live.rs
```diff
@@ -106,10 +106,19 @@ fn get_blocker_engine_deserialized() -> Engine {
     use futures::FutureExt;
     let async_runtime = Runtime::new().expect("Could not start Tokio runtime");
 
+    let brave_service_key = std::env::var("brave_service_key")
```
mihaiplesa, line 109: can we make this uppercase? GitHub only sets them so.

mihaiplesa (changes requested):

tests/live.rs
antonok-edm: Yeah, that's fine too. Was just copying the format from brave-browser `.npmrc`

mihaiplesa: still doesn't find it

Issue #222, opened 2022-06-17
Support redirect priority


antonok-edm:
As per https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#redirect:

> By default, redirect directives have an implicit priority of `0`. Filter authors can declare an explicit priority by appending `:[integer]` (negative values are also supported) to the token of the `redirect=` option, for example:
> 
> ```
> ||example.com/*.js$1p,script,redirect=noopjs:100
> ```
> 
> The priority dictates which redirect token out of many will be ultimately used. Cases of multiple `redirect=` directives applying to a single blocked network request are expected to be rather unlikely. All directives will be reported in the logger, and the effective one will be reported as the last one before redirection entry. Explicit redirect priority should be used if and only if there is a case of redirect ambiguity to solve.

PR #235, opened 2022-11-15
v0.6.0 release

### LINKED ISSUE #166: Support uBO/Adguard removeparam filter
Supporting removeparam, will give us the ability to filter out tracking bits of a url.  

https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#removeparam
https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters?aid=16593#removeparam-modifier

To test in uBO. add `||google.com^$removeparam=/(num|oq|gs_l)/`  or `||google.*^$removeparam=/(num|oq|gs_l)/`

Test the following url: https://www.google.com/search?num=100&q=google&oq=google&gs_l=serp.3..0l10.7976.8565.0.9999.6.6.0.0.0.0.358.617.2-1j1.2.0....0...1c.1.64.serp..4.2.615.BQ3ZvdzuPGE

(In order of codes, expanded also works)
```
$removeparam=num,domain=google.com
$removeparam=oq,domain=google.com
$removeparam=gs_l,domain=google.com
```

The above google.com url simply becomes `https://www.google.com/search?q=google`

**Another sample (with wildcards)**

https://www.bbc.com/news/world-africa-56637573?xtor=AL-72-%5Bpartner%5D-%5Bbbc.news.twitter%5D-%5Bheadline%5D-%5Bnews%5D-%5Bbizdev%5D-%5Bisapi%5D&at_medium=custom7&at_custom2=twitter&at_campaign=64&at_custom3=%40BBCWorld&at_custom1=%5Bpost+type%5D&at_custom4=B2BFEAF6-9605-11EB-97D3-02D74744363C

Can be filtered down too: `||bbc.com^$queryprune=/(xtor.*|at_.*)/`  => https://www.bbc.com/news/world-africa-56637573

A few more filters tested https://0bin.net/paste/vleyxFEC#xMx8TzvreSdTwxUcbVOCFYesnjjXk4wcKCaD-U3bgMR  (on various news sites). Only one bug was attempting to filter `news.google.com` links would cause redirection issues.

antonok-edm:
Of note:
- decoupling of current in-memory representation from last serialized DAT format
- support for `$removeparam` (fixes #166)
- removal of `$bug` option; saving runtime memory on loaded filters
- removal of `csp` field in `NetworkFilter`, saving runtime memory (`$redirect`, `$csp`, and `$removeparam` all share the `modifier_option` field since they cannot be valid at the same time)
- removal of `$redirect-url` support
- `tag` field is now crate-internal, preparing for later removal for further memory savings
- support priorities for `redirect` and `redirect-rule` (fixes https://github.com/brave/adblock-rust/issues/222)


src/blocker.rs
```diff
@@ -265,18 +247,95 @@ impl Blocker {
...
+                    } else {
+                        QParam::KeyOnly(pair)
+                    }
+                })
+                .map(|param| (param, true))
+                .collect();
+
+            let filters = removeparam_filters.check_all(request, &request_tokens, &NO_TAGS);
+            let mut rewrite = false;
+            for removeparam_filter in filters {
```
fmarier, line 311: As an optimization (maybe), we could avoid looping through the params that have been set to `include = false`.

src/blocker.rs
```diff
@@ -1623,6 +1573,175 @@ mod blocker_tests {
         }
     }
 
+    #[test]
+    fn test_removeparam() {
+        let filters = vec![
+            String::from("||example.com^$removeparam=test"),
```
fmarier, line 1579: Should we also test the `mkt_tok` case where that param needs to be removed unless there is an `mkt_unsubscribe` param in the query string? (Or maybe that's not implemented yet? I couldn't find the syntax for it in https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters#removeparam)

src/blocker.rs
antonok-edm: It's doing such a small amount of work per iteration; I'd be more concerned about the ensuing branch predictor misses than the saved cycles. Hard to say without benchmarking of course, but I don't think it's worth it for this small loop :sweat_smile:

src/blocker.rs
antonok-edm: We could do something like `mkt_unsubscribe$removeparam=mkt_tok` - that'd remove `mkt_tok` whenever `mkt_unsubscribe` appeared _anywhere_ in the URL (not just in the query string). The "anywhere" part is probably fine. But I understand the logic is the opposite of what we want, which would be more like:

```adblock
! remove mkt_tok everywhere
*$removeparam=mkt_tok
! *except* when there's an mkt_unsubscribe in the URL
@@mkt_unsubscribe$removeparam=mkt_tok
```

That's gonna require support for removeparam exceptions, which aren't in the uBO default lists and I figured would be best handled in a followup.

src/blocker.rs
fmarier: Yeah, I was just thinking that let's say we're checking `https://example.com/?fbclid=123`, then we'll remove the param as soon as we hit the fbclid rule (i.e. clearing the query string entirely), but we'd still be checking that soon-to-be-removed param against the other `filters`. Seems like we could exit early.

Now that I think more about it though, we could be hitting exception filters later on though that would bring this param back right?

```
||*$removeparam=fbclid
...
@@example.com$removeparam=mkt_tok
```

src/blocker.rs
fmarier: > That'd remove mkt_tok whenever mkt_unsubscribe appeared anywhere in the URL (not just in the query string). The "anywhere" part is probably fine.

Yes, that's also what we settled on in https://github.com/brave/brave-core/pull/13726.

> That's gonna require support for removeparam exceptions, which aren't in the uBO default lists and I figured would be best handled in a followup.

Sounds good. No need to add `mkt_unsubscribe$removeparam=mkt_tok` since as you point out, it's the opposite of what we need anyways. We can add the real one later once exceptions are implemented.

PR #240, opened 2022-12-08
Add a global RegexManager

atuchin-m:
For https://github.com/brave/brave-browser/issues/26892

The idea is to use global regular expression storage instead of individual ones.
This allows to challenge the long-term regexp memory consumption problem in the next PRs.
A new `RegexManager` is in charge of matching/spawning/deleting regex instances. `NetworkFilter` is enforced to match using `RegexManager`.

Plans:
* track the last regex usage and drop the instance it after a few time;
* use less effective, but more memory-saving algorithms for some of the reg exps until it becomes a hot path;

Note:
* Using a standard hasher for u64 instead of AHasher results in 3% performance drop.. Not sure AHasher is the best option. I am open to other suggestions.
* Replacing a standard hasher in other places could also give a perf boost, worst trying.
* `get_debug_info()` method is supposed to be exposed to something like `brave://adblock-internals`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ exclude = [
 ]
 
 [dependencies]
-ahash = "0.8.0"
 addr = { version = "0.14", default-features = false, features = ["psl"], optional = true }
 url = "2.2"
 percent-encoding = "2.1"
@@ -87,11 +86,12 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling"]
+default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "unsync-regex-caching"]
 metrics = []
 full-regex-handling = []
 object-pooling = ["lifeguard"] # disables `Send` and `Sync` on `Engine`.
 unsync-regex-caching = [] # disables `Send` and `Sync` on `Engine`.
+debug-info = []
 css-validation = ["cssparser", "selectors"]
 content-blocking = ["serde_json"]
 embedded-domain-resolver = ["addr"] # Requires setting an external domain resolver if disabled.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ bitflags = "1.2"
 itertools = "0.10"
 idna = "0.2"
 serde =  { version = "1.0", features = ["derive", "rc"] }
-mock_instant = { version = "0.2", features = ["sync"] }
 flate2 = { version = "1.0", features = ["rust_backend"], default-features = false }
 seahash = "3"   # seahash 4 introduces a breaking hash algorithm change
 twoway = "0.2"
@@ -48,6 +47,7 @@ serde_json = { version = "1.0", optional = true }
 [dev-dependencies]
 criterion = "0.3"
 csv = "1"
+mock_instant = { version = "0.2", features = ["sync"] }
 serde_json = "1.0"
 # By default, reqwest builds openssl from source, which fails on missing/incompatible system dependencies
 reqwest = { version = "0.11", features = ["rustls-tls"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,7 @@ default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling"]
 metrics = []
 full-regex-handling = []
 object-pooling = ["lifeguard"] # disables `Send` and `Sync` on `Engine`.
+unsync-regex-caching = [] # disables `Send` and `Sync` on `Engine`.
 css-validation = ["cssparser", "selectors"]
 content-blocking = ["serde_json"]
 embedded-domain-resolver = ["addr"] # Requires setting an external domain resolver if disabled.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ exclude = [
 ]
 
 [dependencies]
+ahash = "0.8.0"
 addr = { version = "0.14", default-features = false, features = ["psl"], optional = true }
 url = "2.2"
 percent-encoding = "2.1"
@@ -86,11 +87,10 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "unsync-regex-caching"]
+default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling"]
 metrics = []
 full-regex-handling = []
 object-pooling = ["lifeguard"] # disables `Send` and `Sync` on `Engine`.
-unsync-regex-caching = [] # disables `Send` and `Sync` on `Engine`.
 css-validation = ["cssparser", "selectors"]
 content-blocking = ["serde_json"]
 embedded-domain-resolver = ["addr"] # Requires setting an external domain resolver if disabled.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ bitflags = "1.2"
 itertools = "0.10"
 idna = "0.2"
 serde =  { version = "1.0", features = ["derive", "rc"] }
+mock_instant = { version = "0.2", features = ["sync"] }
 flate2 = { version = "1.0", features = ["rust_backend"], default-features = false }
 seahash = "3"   # seahash 4 introduces a breaking hash algorithm change
 twoway = "0.2"
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -1,8 +1,6 @@
 //! Holds `Blocker`, which handles all network-based adblocking queries.
 
 use once_cell::sync::Lazy;
-use std::borrow::BorrowMut;
-use std::cell::RefCell;
 use std::ops::DerefMut;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
@@ -137,7 +135,10 @@ pub struct Blocker {
     pub(crate) pool: TokenPool,
 
     // Not serialized
-    pub(crate) regex_manager: RefCell<RegexManager>,
+    #[cfg(feature = "unsync-regex-caching")]
+    pub(crate) regex_manager: std::cell::RefCell<RegexManager>,
+    #[cfg(not(feature = "unsync-regex-caching"))]
+    pub(crate) regex_manager: std::sync::Mutex<RegexManager>,
 }
 
 impl Blocker {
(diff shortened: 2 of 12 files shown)


antonok-edm (commented):
I think overall this approach is a good idea.

I noticed a few places where the indentation is a bit messed up. I'd recommend `cargo fmt`, but currently it would produce a bunch of unrelated changes, which is not ideal. Could you at least go over the parts you've touched and make sure the indentation is 4 spaces? If it's easier you can copy your code changes into another file and format that one specifically.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ exclude = [
 ]
 
 [dependencies]
+ahash = "0.8.0"
```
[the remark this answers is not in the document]
atuchin-m: Thanks, I will compare. I believe we don't need 3 different hash function (the standard, seahash, ahash) ;)

src/blocker.rs
```diff
@@ -634,6 +650,15 @@ impl Blocker {
...
         self.resources.get_resource(key)
     }
+
+    pub fn get_debug_info(&self) -> BlockerDebugInfo {
+      let regex_manager = self.regex_manager.borrow();
+      BlockerDebugInfo{
+        active_regex_count: regex_manager.get_active_regex_count(),
+        compiled_regex_count: regex_manager.get_compiled_regex_count()
+      }
+    }
```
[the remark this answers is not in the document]
atuchin-m: Ok.

src/filters/network.rs
```diff
@@ -1097,32 +1098,20 @@ impl fmt::Display for NetworkFilter {
 }
 
 pub trait NetworkMatchable {
-    fn matches(&self, request: &request::Request) -> bool;
-    fn get_regex(&self) -> Arc<CompiledRegex>;
+    fn matches(&self, request: &request::Request, regex_manager: &mut RegexManager) -> bool;
+
+    #[cfg(test)]
+    fn matches_test(&self, request: &request::Request) -> bool;
```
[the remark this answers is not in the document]
atuchin-m: The idea was to force all the instances to use a correct `RegexManager` (except the tests).
The thing that I'm afraid that we occasionally call the wrong function in the production code. 
Like returning in old signature `fn matches(&self, request: &request::Request) -> bool ` results in still usin  it in some part of adblock-rust or 3rd party product.

That's why I believe we'd better to make breaking changes here.
A singleton as alternative doesn't sounds good..

atuchin-m:
@antonok-edm PTAL

src/blocker.rs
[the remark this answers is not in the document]
atuchin-m: Done.

Cargo.toml
```diff
@@ -86,11 +87,10 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "unsync-regex-caching"]
+default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling"]
 metrics = []
 full-regex-handling = []
 object-pooling = ["lifeguard"] # disables `Send` and `Sync` on `Engine`.
-unsync-regex-caching = [] # disables `Send` and `Sync` on `Engine`.
```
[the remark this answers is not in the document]
atuchin-m: Done.

Cargo.toml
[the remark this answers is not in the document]
atuchin-m: Replaced to seahash. ahash a little bit faster but it not worse to add an extra dependency.

src/filters/network.rs
```diff
@@ -272,6 +274,54 @@ impl NetworkFilterOption {
...
+    }
+  }
+}
+
+impl RegexManager {
+  pub fn matches(&mut self, filter: &NetworkFilter, pattern: &str) -> bool {
+      if !filter.is_regex() && !filter.is_complete_regex() {
+          return true;
+      }
+      let key = filter as *const NetworkFilter as usize;
```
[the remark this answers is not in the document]
atuchin-m: Done.

src/filters/network.rs
```diff
@@ -3283,17 +3329,11 @@ mod match_tests {
         {
             let filter = r#"/^https?:\/\/([0-9a-z\-]+\.)?(9anime|animeland|animenova|animeplus|animetoon|animewow|gamestorrent|goodanime|gogoanime|igg-games|kimcartoon|memecenter|readcomiconline|toonget|toonova|watchcartoononline)\.[a-z]{2,4}\/(?!([Ee]xternal|[Ii]mages|[Ss]cripts|[Uu]ploads|ac|ajax|assets|combined|content|cov|cover|(img\/bg)|(img\/icon)|inc|jwplayer|player|playlist-cat-rss|static|thumbs|wp-content|wp-includes)\/)(.*)/$image,other,script,~third-party,xmlhttprequest,domain=~animeland.hu"#;
             let network_filter = NetworkFilter::parse(filter, true, Default::default()).unwrap();
-            let regex = Arc::try_unwrap(network_filter.get_regex()).unwrap();
-            assert!(
```
atuchin-m, line 3287: Removed because there is no way to get the `CompiledRegex` type.
Do it important here to check it?

antonok-edm: This test is just checking that the filter has a compiled regex. The equivalent would be checking that an external `RegexManager` compiles the regex for it (after a `matches` call). I don't think it's critical in this test, but we probably should have at least one test somewhere to confirm that the regex count increases as expected.

antonok-edm (commented):
I think it would be good to break `RegexManager` out into its own file; just another top-level `regex_manager.rs` would be fine. If you can do that and add a test or two there, I think it should be good to go.

src/filters/network.rs
```diff
@@ -151,13 +152,15 @@ impl From<&request::RequestType> for NetworkFilterMask {
 pub enum CompiledRegex {
     Compiled(Regex),
     CompiledSet(RegexSet),
+    None,
```
atuchin-m, line 155: I've added this new entry instead of using `Option<CompiledRegex>`.
This allow to avoid growing the map entry to +8 bytes (because of extra `Option`).

Motivation:  https://adeschamps.github.io/enum-size

src/filters/network.rs
atuchin-m: ok, I see.
I've modified the test to check `get_compiled_regex_count()` before/after calling `is_match`.

src/filters/network.rs
```diff
@@ -272,6 +272,104 @@ impl NetworkFilterOption {
...
+        }
+    }
+
+    pub fn cleanup(&mut self) {
+        self.map.clear();
+        let now = self.now;
+        self.map.retain(|_, v| now - v.last_used < std::time::Duration::from_secs(180));
+        for (_, v) in &self.map {
+          println!("{}", v.regex)
+        }
```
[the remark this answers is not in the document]
atuchin-m: Thanks for the correction.
Originally, that code was supposed to be in the next PR, but after the plan was changed.
So I messed up the branches and some test code went here.

Now I've changed the solution to preserve the map entries but to drop the regexs themselves.
This gives extra information about which reg expressions were discarded.

src/filters/network.rs
```diff
@@ -152,13 +152,15 @@ impl From<&request::RequestType> for NetworkFilterMask {
...
     CompiledSet(RegexSet),
+    None,
     MatchAll,
     RegexParsingError(regex::Error),
 }
 
 impl CompiledRegex {
     pub fn is_match(&self, pattern: &str) -> bool {
         match &self {
+            CompiledRegex::None => {panic!("use of CompiledRegex::None")} // No value, invalid call.
```
antonok-edm, line 163: This is the kind of thing I'd prefer to see behind the `debug-info` flag, so we can guarantee it won't cause crashes for users in the wild.

Can we do this?
- Rename `CompiledRegex::None` to `CompiledRegex::Discarded`
- Put `CompiledRegex::Discarded` behind a `#[cfg(feature = "debug-info")]`
- Completely remove the `CompiledRegex` in `cleanup` when `debug-info` is disabled

src/filters/regex_manager.rs
```diff
@@ -87,14 +96,14 @@ impl RegexManager {
...
-            if now - v.last_used < std::time::Duration::from_secs(180) {
+            if now - v.last_used >= REGEX_MANAGER_DISCARD_TIME {
                 // Discard the regex to save memory.
                 v.regex = CompiledRegex::None;
             }
         }
     }
 
-    #[cfg(feature = "debug-info")]
+    #[cfg(any(feature = "debug-info", test))]
```
antonok-edm, line 106: we can leave these as `#[cfg(feature = "debug-info")]` - just make `mod tests` also `#[cfg(feature = "debug-info")]`.

src/filters/network.rs
atuchin-m: In theory we can, I also though about the similar approach.
But I believe we need discarded map entries to make a better decision in future (=stop discarding the entries that we often recreate.

src/filters/regex_manager.rs
atuchin-m: If we could suppose that `debug-info` is always enable for tests then it's okay.
I worried about the test coverage for configurations with !debug-info.

atuchin-m:
@antonok-edm Any other comments?

src/filters/network.rs
antonok-edm: > But I believe we need discarded map entries to make a better decision in future (=stop discarding the entries that we often recreate.

I think that makes sense as an approach, but I'd strongly prefer to add panics like this only when necessary. Unless you're planning to implement that strategy in this PR, let's leave it out and we can revisit later.

There could also be alternatives to this pattern that wouldn't panic (maybe we can store hashes of the discarded rules in a bloom filter or hashmap?). But we should look at that holistically.

src/filters/regex_manager.rs
antonok-edm: Yes, CI runs with all features enabled

atuchin-m: Fixed.

src/filters/network.rs
atuchin-m: Changed to `Option<CompiledRegex>`.

src/filters/regex_manager.rs
```diff
@@ -127,7 +127,7 @@ impl RegexManager {
     }
 }
 
-#[cfg(test)]
+#[cfg(feature = "debug-info")]
```
antonok-edm, line 130: should be `#[cfg(all(test, feature = "debug-info"))]`

src/filters/regex_manager.rs
```diff
@@ -144,13 +144,7 @@ mod tests {
...
             .get_debug_regex_data()
             .iter()
-            .fold(0, |acc, x| {
-                if x.regex == CompiledRegex::None.to_string() {
-                    acc
-                } else {
-                    acc + 1
-                }
-            })
+            .fold(0, |acc, x| if x.regex.is_some() { acc + 1 } else { acc })
```
antonok-edm, line 149: would be better expressed with a `filter` operation:

```rust
.filter(|x| x.regex.is_some())
.count()
```

Cargo.toml
```diff
@@ -33,6 +33,7 @@ bitflags = "1.2"
 itertools = "0.10"
 idna = "0.2"
 serde =  { version = "1.0", features = ["derive", "rc"] }
+mock_instant = { version = "0.2", features = ["sync"] }
```
antonok-edm, line 36: this should be in `dev-dependencies`

antonok-edm (commented):
last few comments and it should be good to go

src/filters/regex_manager.rs
atuchin-m: Fixed

Cargo.toml
atuchin-m: Fixed

src/filters/regex_manager.rs
atuchin-m: Done

Issue #242, opened 2022-12-19
Support for better regex filters.


ryanbr:
Used in uBO, https://github.com/uBlockOrigin/uAssets/blob/master/filters/badware.txt#L881
https://github.com/uBlockOrigin/uAssets/blob/master/filters/filters.txt#L19296

The regex filters used for revolving adserver domains.  From what I see, they are ignored in Brave

```
badware.txt:/^https:\/\/[0-9a-z]{3,}\.[-a-z]{10,}\.(?:li[fv]e|top|xyz)\/[a-z]{8}\/\?utm_campaign=\w{40,}/$doc,match-case,domain=life|live|top|xyz
badware.txt:/^https?:\/\/((?!www)[a-z]{3,}|\d{2})?\.?[-0-9a-z]{6,}\.[a-z]{2,6}\/(?:[a-z]{6,8}\/)?\/?\?u=[0-9a-z]{7}&o=[0-9a-z]{7}/$doc,frame,match-case,domain=buzz|com|de|fun|guru|info|life|live|mobi|online|pw|site|space|top|us|xyz
badware.txt:/^https?:\/\/((?!www)[a-z]{3,}|\d{2})?\.?[-0-9a-z]{6,}\.[a-z]{2,6}\/(?:[a-z]{6,8}\/)?\/?\?o=[0-9a-z]{7}&u=[0-9a-z]{7}/$doc,frame,match-case,domain=buzz|com|de|fun|guru|info|life|live|mobi|online|pw|site|space|top|us|xyz
badware.txt:/^https:\/\/(?:www\d\.)?[-a-z]{6,}\.(?:com|info|net|org)\/(?=[-_a-zA-Z]{0,42}\d)(?=[-_0-9a-z]{0,42}[A-Z])[-_0-9a-zA-Z]{43}\/\?cid=[-_0-9a-zA-Z]{16,36}(?:&qs\d=\S+)?&sid=[_0-9a-f]{1,32}$/$doc,match-case,domain=com|info|net|org
badware.txt:/^https:\/\/(?:www\d\.)?[-a-z]{6,}\.(?:com|info|net|org)\/(?=[-_a-zA-Z]{0,42}\d)(?=[-_0-9a-z]{0,42}[A-Z])[-_0-9a-zA-Z]{43}\/\?sid=[_0-9a-f]{1,32}(?:&qs\d=\S+)?&cid=[-_0-9a-zA-Z]{16,36}$/$doc,match-case,domain=com|info|net|org
[...]

PR #243, opened 2023-01-05
Support for better match-case regex filters

ankeleralph:
This PR fixes issue #242 Support for better match-case regex filters.

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -1269,10 +1269,10 @@ fn get_url_after_hostname<'a>(url: &'a str, hostname: &str) -> &'a str {
 fn check_pattern_plain_filter_filter(filter: &NetworkFilter, request: &request::Request) -> bool {
     match &filter.filter {
         FilterPart::Empty => true,
-        FilterPart::Simple(f) => twoway::find_str(&request.url, f).is_some(),
+        FilterPart::Simple(f) => twoway::find_str(if filter.match_case() {&request.original_url} else {&request.url}, f).is_some(),
         FilterPart::AnyOf(filters) => {
             for f in filters {
-                if twoway::find_str(&request.url, f).is_some() {
+                if twoway::find_str(if filter.match_case() {&request.original_url} else {&request.url}, f).is_some() {
                     return true;
                 }
             }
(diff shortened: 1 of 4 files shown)


src/filters/network.rs
```diff
@@ -1351,11 +1361,17 @@ fn check_pattern_right_anchor_filter(filter: &NetworkFilter, request: &request::
...
-                if request.url.starts_with(f) {
-                    return true;
+                if filter.match_case() {
+                    if request.original_url.starts_with(f) {
+                        return true;
+                    }
+                } else {
+                    if request.url.starts_with(f) {
+                        return true;
+                    }
```
antonok-edm, line 1374: for patterns like this where both branches are the same except for the URL to check, it'd be easier to understand if you store a reference to the correct URL first, i.e.

```rust
let request_url = if filter.match_case() {
    &request.original_url
} else {
    &request.url
};
if request_url.starts_with(f) {
    return true;
}
```

src/filters/network.rs
```diff
@@ -1466,9 +1502,9 @@ fn check_pattern_hostname_left_right_anchor_filter(
                     // Since it must follow immediatly after the hostname and be a suffix of
                     // the URL, we conclude that filter must be equal to the part of the
                     // url following the hostname.
-                    FilterPart::Simple(f) => get_url_after_hostname(&request.url, hostname) == f,
+                    FilterPart::Simple(f) => get_url_after_hostname(if filter.match_case() {&request.original_url} else {&request.url}, hostname) == f,
                     FilterPart::AnyOf(filters) => {
-                        let url_after_hostname = get_url_after_hostname(&request.url, hostname);
+                        let url_after_hostname = get_url_after_hostname(if filter.match_case() {&request.original_url} else {&request.url}, hostname);
```
antonok-edm, line 1507: it'd also help to pull `request_url` out earlier in cases like this too, rather than checking `filter.match_case()` multiple times.

```rust
let request_url = if filter.match_case() {
    &request.original_url
} else {
    &request.url
};
//...
    FilterPart::Simple(f) => get_url_after_hostname(request_url, hostname) == f,
//...
        let url_after_hostname = get_url_after_hostname(request_url, hostname);
```

src/filters/network.rs
ankeleralph: Thank you. Fixed.

src/filters/network.rs
ankeleralph: Thank you. Fixed.

src/filters/network.rs
```diff
@@ -1368,12 +1387,19 @@ fn check_pattern_left_right_anchor_filter(
...
+        &request.original_url
+    } else {
+        &request.url
+    };
     match &filter.filter {
         FilterPart::Empty => true,
-        FilterPart::Simple(f) => &request.url == f,
+        FilterPart::Simple(f) => {
+            return request_url == f;
+        },
```
antonok-edm, line 1399: original single-line syntax here would be preferred

antonok-edm:
A few changes:
- rearrange places where `Request::url` was lowercased (there was a conflict with `original_url` which wasn't punycoded, yet we were taking the same indices into it as from the other `url` field)
- prevented `$match-case` on non-regex filters (see https://github.com/gorhill/ublock/wiki/Static-filter-syntax#match-case)
- noted that some tests fail due to regex filters of the form `/...$/$...` (the first `$` should be treated as the end of the URL, not the pattern/options separator)
- a few tests for unicode, and updated other tests to match

PR #254, opened 2023-01-31
Fusion optimization now able to return regular patterns

atuchin-m:
Resolves https://github.com/brave/adblock-rust/issues/255

image

### DIFF
--- a/src/optimizer.rs
+++ b/src/optimizer.rs
@@ -102,8 +102,7 @@ impl Optimization for SimplePatternGroup {
             }
         }
 
-        // let is_regex = filters.iter().find(|f| f.is_regex()).is_some();
-        let is_regex = true;
+        let is_regex = filters.iter().find(|f| f.is_regex()).is_some();
         filter.mask.set(NetworkFilterMask::IS_REGEX, is_regex);
         let is_complete_regex = filters.iter().any(|f| f.is_complete_regex());
         filter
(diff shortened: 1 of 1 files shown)


src/optimizer.rs
```diff
@@ -222,15 +221,19 @@ mod optimization_tests_pattern_group {
     fn check_match(
```
atuchin-m, line 221: check_match should take into account the filter could be non-regex.

src/optimizer.rs
```diff
@@ -284,7 +287,7 @@ mod optimization_tests_pattern_group {
 
         let fused = optimization.fusion(&filters);
 
-        assert!(fused.is_regex(), "Expected rule to be regex");
+        assert!(fused.is_regex() == false, "Expected rule to not be a regex");
```
atuchin-m, line 290: Now it checks non-regex case. We could also add another case if needed.

src/optimizer.rs
```diff
@@ -102,8 +102,7 @@ impl Optimization for SimplePatternGroup {
             }
         }
 
-        // let is_regex = filters.iter().find(|f| f.is_regex()).is_some();
```
atuchin-m, line 105: Any idea why it was disabled?
I also believe we shouldn't fuse regex with non regex rules.
So a possible solution is returning a pair or a vector `[fused_non_regex_rules, fused_regex_rules]`

antonok-edm: https://github.com/brave/adblock-rust/commit/5c28aefc#diff-40a35a91a642ade2cf6b8ce86ddb3c8b82625c0592d0e9fa24d2b41a072ab3adR92

No idea :stuck_out_tongue:

src/optimizer.rs
antonok-edm: Sounds like there are gonna be some changes here anyways so this should be fine for now

antonok-edm (approved):

PR #261, opened 2023-02-01
`RegexManager` is `Send`

### LINKED ISSUE #260: `Engine` struct is no longer `core::marker::Sync`
The PyO3 library requires structures to be `Sync`. This means that [`python-adblock`](https://github.com/ArniDagur/python-adblock) requires `Engine` to be `Sync`. Previously, this was made possible by disabling the `object-pooling` feature. However, there has been a regression, and `Engine` is unconditionally `Sync`.

The immediate cause is that there is a `*const NetworkFilter` pointer in the `RegexManager` structure. This can be fixed by applying the following patch:
```diff
diff --git a/src/regex_manager.rs b/src/regex_manager.rs
index 7217086..4e86c5a 100644
--- a/src/regex_manager.rs
+++ b/src/regex_manager.rs
@@ -57,7 +57,7 @@ impl Default for RegexManagerDiscardPolicy {
 type RandomState = std::hash::BuildHasherDefault<seahash::SeaHasher>;
 
 pub struct RegexManager {
-    map: HashMap<*const NetworkFilter, RegexEntry, RandomState>,
+    map: HashMap<usize, RegexEntry, RandomState>,
     compiled_regex_count: usize,
     now: Instant,
     last_cleanup: Instant,
@@ -90,7 +90,10 @@ impl RegexManager {
         if !filter.is_regex() && !filter.is_complete_regex() {
             return true;
         }
-        let key = filter as *const NetworkFilter;
+        // Converting to `usize` prevents `*const NetworkFilter` appearing in
+        // the struct definition of `RegexManager`, which keeps `RegexManager`
+        // `core::marker::Sync`.
+        let key = filter as *const NetworkFilter as usize;
         use std::collections::hash_map::Entry;
         match self.map.entry(key) {
             Entry::Occupied(mut e) => {
```

There may be more places that need fixing, but I haven't found out yet because of the issue described in https://github.com/brave/adblock-rust/pull/259.

antonok-edm:
Should fix #260

cc @ArniDagur

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,3 +16,17 @@ pub mod resources;
 pub mod url_parser;
 #[doc(hidden)]
 pub mod utils;
+
+#[cfg(test)]
+mod sync_tests {
+    #[allow(unused)]
+    fn static_assert_sync<S: Sync>() {
+        let _ = core::marker::PhantomData::<S>::default();
+    }
+
+    #[test]
+    #[cfg(not(any(feature = "object-pooling", feature = "unsync-regex-caching")))]
+    fn assert_engine_sync() {
+        static_assert_sync::<crate::engine::Engine>();
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,4 @@
 #![allow(dead_code)]
-#![forbid(unsafe_code)]
 
 // Own modules, currently everything is exposed, will need to limit
 pub mod blocker;
--- a/src/regex_manager.rs
+++ b/src/regex_manager.rs
@@ -24,6 +24,10 @@ impl Instant {
     }
 }
 
+/// `*const NetworkFilter` could technically leak across threads through `RegexDebugEntry::id`, but
+/// it's disguised as a unique identifier and not intended to be dereferenced.
+unsafe impl Send for RegexManager {}
+
 const DEFAULT_CLEAN_UP_INTERVAL: Duration = Duration::from_secs(30);
 const DEFAULT_DISCARD_UNUSED_TIME: Duration = Duration::from_secs(180);
 


src/regex_manager.rs
```diff
@@ -90,7 +90,7 @@ impl RegexManager {
         if !filter.is_regex() && !filter.is_complete_regex() {
             return true;
         }
-        let key = filter as *const NetworkFilter;
+        let key = filter.get_id();
```
atuchin-m, line 93: Could we relay on ids are unique?

src/regex_manager.rs
```diff
@@ -90,7 +90,7 @@ impl RegexManager {
         if !filter.is_regex() && !filter.is_complete_regex() {
             return true;
         }
-        let key = filter as *const NetworkFilter;
```
atuchin-m, line 93: Im fact we inside the lock (for the multiple threads case). Frankly I'm don't fully understand what's wrong with taking the pointer.

atuchin-m:
Also `get_id()` looks a super expensive in terms of CPU time.

antonok-edm:
I investigated a bit further.

Rust's `Mutex<T>` will become `Sync` if `T` is `Send`. `RegexManager` currently is _not_ `Send`, so `Mutex<RegexManager>` not becoming `Sync` either.

(For reference: `Send` = safe to send to another thread, `Sync` = safe to share between multiple threads)

`RegexManager` is not `Send` because it deals with `*const NetworkFilter`. The concern is that any dangling pointers must all be moved to the new thread with the `RegexManager`, and this is not possible for Rust's safety guards to guarantee.

Our specific use of `RegexManager` does _technically_ leak pointers in `RegexManager::get_debug_regex_data`, although they're disguised as `u64` unique identifiers. Somebody could cast those identifiers back to `*const NetworkFilter`s again if they _really_ wanted to, but actually dereferencing those pointers will require `unsafe` code anyways. So I think it's fine to manually mark `RegexManager` as `Send`.

src/lib.rs
```diff
@@ -1,5 +1,4 @@
 #![allow(dead_code)]
-#![forbid(unsafe_code)]
```
antonok-edm, line 2: :smiling_face_with_tear:

src/lib.rs
```diff
@@ -17,3 +16,17 @@ pub mod resources;
...
+    #[allow(unused)]
+    fn static_assert_sync<S: Sync>() {
+        let _ = core::marker::PhantomData::<S>::default();
+    }
+
+    #[test]
+    #[cfg(not(any(feature = "object-pooling", feature = "unsync-regex-caching")))]
+    fn assert_engine_sync() {
+        static_assert_sync::<crate::engine::Engine>();
+    }
```
antonok-edm, line 31: @ArniDagur I added a test to make sure this doesn't regress again

src/regex_manager.rs
```diff
@@ -24,6 +24,10 @@ impl Instant {
     }
 }
 
+/// `*const NetworkFilter` could technically leak across threads through `RegexDebugEntry::id`, but
+/// it's disguised as a unique identifier and not intended to be dereferenced.
```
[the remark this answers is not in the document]
atuchin-m: Let me clarify:
1. no change for `map` property
2. change the way how we  create a `RegexDebugEntry` from a map entry.

PR #263, opened 2023-02-02
Bump criterion to 0.4

rillian:
Update the criterion benchmarking utility to 0.4 to address audit warnings about the deprecated cbor crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,6 +60,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


rillian:
NB `cargo bench` panics for me with or without this patch, but I believe the benchmark api we're using is the same, so hopefully it's not hiding any new issues.

antonok-edm:
Yes, the panic is unrelated. Thanks for helping with this!

Issue #262, opened 2023-02-02
twoway crate is unmaintained


rillian:
The `twoway` crate used for fast string matching was recently marked as unmaintained. It would be good to migrate away to avoid triggering audit warnings.

The (former) maintainers recommend using `memchr` but at first glance that supports byte slices but has no equivalent to `twoway::find_str`, so in practice it may need to be some combination of `std::str::find` and `mem(r)chr` with byte string literals and `.as_bytes()` whichever benchmarks better.

wknapik:
When closing this, please also close https://github.com/brave/brave-browser/issues/28228 (but not before the fix is present on all release branches, or the issue will be automatically reopened).

antonok-edm:
Just noting that `memchr` does have [`memmem`](https://docs.rs/memchr/latest/memchr/memmem/index.html) for substring find.

rillian:
:pray:

PR #264, opened 2023-02-03
switch from `twoway` to `memchr` and replace existing stdlib methods

antonok-edm:
Closes https://github.com/brave/adblock-rust/issues/262

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,6 +17,7 @@ dependencies = [
  "idna 0.2.3",
  "itertools",
  "lifeguard",
+ "memchr",
  "mock_instant",
  "once_cell",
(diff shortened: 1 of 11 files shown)


antonok-edm:
wow, who knew that a few `unwrap`s with documented safety invariants would tag the whole security team :sweat_smile:

Is there a lint option I can add to ignore the warnings? I totally see the value in this for code that's accidentally committed, but it'd be nice to have a way to say "yes, I understand the consequences here"

thypon:
It's fine ahah. You can add a '// nosemgrep <rule-id>' either the line before, or the line after. RuleID is typically the string that is after the last dot in "source"

PR #275, opened 2023-02-24
Implement Error/Display for error types, implement From<&str> for MimeType

DJAndries:


src/filters/network.rs
```diff
@@ -15,31 +16,55 @@ pub const TOKENS_BUFFER_SIZE: usize = 200;
...
+    #[error("full regex unsupported")]
     FullRegexUnsupported,
+    #[error("regex parsing error")]
     RegexParsingError(regex::Error),
+    #[error("punycode error")]
     PunycodeError,
+    #[error("csp with content type")]
     CspWithContentType,
+    #[error("match case without full regex")]
     MatchCaseWithoutFullRegex,
```
antonok-edm, line 68: `match-case without full regex`

antonok-edm (commented):
Sorry for the delay! Just a few nits - some of these are referring to actual filter syntax, and it'd be good to have those keywords spelled out correctly in the error messages.

antonok-edm (approved):

antonok-edm:
Awesome, thanks!

PR #291, opened 2023-04-30
Expose cosmetic filtering methods to JS bindings

### LINKED ISSUE #290: Cosmetic filtering via JS binding
I would like to find ad elements in pages in JS.

I just tried using the `adblock-rs`, and then I found that it can be determined by the link of the image, but not by the cosmetic filter.

It would be nice if the ability to detect advertising elements from a page with a cosmetic filter could also be used from Javascript.

antonok-edm:
Closes #290 (cc @eggplants)

I'd still like to experiment with making `exceptions` an opaque struct to avoid serialization/deserialization overhead, but this should otherwise be functional if you'd like to start testing it.

```js
const urlCosmeticResources = engine.urlCosmeticResources('https://youtube.com');
const hiddenClassIdSelectors = engine.hiddenClassIdSelectors(['ad'], [], urlCosmeticResources.exceptions);
```

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -22,7 +22,7 @@ class Engine {
         this.boxed = blocker.Engine_constructor(filter_set.boxed, ...args);
     }
 }
-forwardClassMethods(Engine, ['check', 'serializeRaw', 'serializeCompressed', 'deserialize', 'enableTag', 'useResources', 'tagExists', 'clearTags', 'addResource', 'getResource']);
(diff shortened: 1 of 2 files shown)


antonok-edm:
After further experimentation, making `exceptions` opaque doesn't seem to be worth the additional code maintenance. The full set of exceptions for any given site is already going to be quite small, so serialization/deserialization overhead is minimal to begin with. And who knows, maybe somebody will find it useful to be able to introspect the exceptions.

I'll go ahead and merge this and push a new release with it.

PR #298, opened 2023-07-04
Increase permissiveness when parsing the `Expires` special comment

DJAndries:
It seems that there are quite a few lists that have the suffix ` (update frequency)` included in the `Expires` special comment. `ExpiresInterval::try_from` fails if this suffix exists. If this is intended, feel free to close this PR.

This also increases permissiveness for the plurality of the time unit; i.e. `1 hours` and `2 hour` would be acceptable values. This adjustment can be removed if need be.


antonok-edm (approved):
I tried to follow [the original spec](https://help.adblockplus.org/hc/en-us/articles/360062733293#special-comments) as closely as possible, although it's not particularly descriptive and I've noticed more of these cases in the wild. Thanks @DJAndries!

Issue #310, opened 2023-09-11
Support single-quoted argument syntax


antonok-edm:
As of https://github.com/gorhill/ublock/commit/fd036a51, uBO supports single-quoted arguments in scriptlet injections to avoid having to escape commas. adblock-rust should support that as well. See #307 for additional context.

Emi-Emi-Emi:
This is working great in latest Nightly, thanks for the fix/compatibility support @antonok-edm. 👍

It is good now all Scriptlet Injections will work as expected. No more pages breaking because of this, especially Youtube, since uBlock Quick-Fixes added a week, ago again, for many days (and then removed) a similar rule like the one I reported when I requested this in the other issue, and doing the same breakage where Youtube wouldn't load since the rule couldn't do the job properly.
https://github.com/uBlockOrigin/uAssets/commit/50819e37542a8a080f03ea987e3a77f35ee0a1de

Also many of rules in uBlock annoyances/cookies contained this quoted syntax even if they didn't need it and had no commas to escape, rendering the rule useless for no reason, so, this is great.

PR #323, opened 2023-11-07
Support quoted scriptlet arguments

antonok-edm:


antonok-edm:
(closed https://github.com/brave/adblock-rust/issues/310)

PR #337, opened 2024-02-21
Don't crash when converting content blocking rules with $from option

### LINKED ISSUE #336: Missing `from` support on `content_blocking.rs` causes crashes
```
thread '<unnamed>' panicked at ../../brave/third_party/rust/adblock/v0_8/crate/src/content_blocking.rs:407:69:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

ContentBlockerManager.swift:176:
func compileRuleList(at localFileURL: URL, for type: BlocklistType, modes: [BlockingMode]) async throws {

   //See the attached file `FilterSet-ContentBlocker-Manager.txt` for `filterSet` String.
}

adblock_engine.mm:235:
+ (nullable ContentBlockingRulesResult*)
    contentBlockerRulesFromFilterSet:(NSString*)filterSet
                               error:(NSError**)error {
  auto result = adblock::convert_rules_to_content_blocking(
      base::SysNSStringToUTF8(filterSet));
  
   //See the attached file `Filter-Set-Adblock_Engine-mm.txt` for the `filterSet` NSString
}

lists.rs:277:
self.network_filters.into_iter().for_each(|filter| {
            let original_rule = *filter.raw_line.clone().expect("All rules should be in debug mode");
            if let Ok(equivalent) = TryInto::<content_blocking::CbRuleEquivalent>::try_into(filter) {
                filters_used.push(original_rule);
                equivalent.into_iter().for_each(|cb_rule| {
                    match &cb_rule.action.typ {
                        content_blocking::CbType::IgnorePreviousRules => ignore_previous_rules.push(cb_rule),
                        _ => other_rules.push(cb_rule),
                    }
                });
            }
        });

Printing description of original_rule:
(alloc::string::String) original_rule = {
  vec = {
    buf = {
      ptr = {
        pointer = (pointer = "@@||cmp.pubtech.ai/*/pubtech-cmp-v2.js$script,from=ilrestodelcarlino.it|quotidiano.net|lanazione.it|ilgiorno.it|iltelegrafolivorno.it")
        _marker = {}
      }
      cap = 133
      alloc = {}
    }
    len = 133
  }


Printing description of opts.data_ptr:
(unsigned char *) data_ptr = 0x000060000305ddf7 "script,from=ilrestodelcarlino.it|quotidiano.net|lanazione.it|ilgiorno.it|iltelegrafolivorno.it"
```

It seems like from alias support was added here: https://github.com/brave/adblock-rust/commit/1f966be5586bb09a3a02219478cafd440b7e77f0
But the corresponding content_blocker rule conversion only checks for domain: https://github.com/brave/adblock-rust/blob/f3c7fbbf0b98fabbdbcf076bffaeee187a701465/src/content_blocking.rs#L405

antonok-edm:
Fixes #336.

Just ignoring the rules for now; the code can be improved to accept them later.

### DIFF
--- a/src/content_blocking.rs
+++ b/src/content_blocking.rs
@@ -202,6 +202,8 @@ pub enum CbRuleCreationFailure {
     ScriptletInjectionsNotSupported,
     /// Valid content blocking rules can only include ASCII characters.
     RuleContainsNonASCII,
+    /// `from` as a `domain` alias is not currently supported in content blocking syntax.
+    FromNotSupported,
 }
 
 impl TryFrom<ParsedFilter> for CbRuleEquivalent {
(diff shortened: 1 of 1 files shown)


cuba (approved):
Ran against an iOS conversion test and passes

Issue #340, opened 2024-03-07
Wrong types for js bindings


zack-esparza:
## request_type

Current:
```ts
request_type: string
```

Expected: 

```ts
request_type: RequestType // union type or enum
```
## engine.check

Current: 

```ts
check(url: string, source_url: string, request_type: string, debug?: boolean): boolean;
```

Excpeted: 
```
check(url: string, source_url: string, request_type: string, debug?: boolean): boolean | DebugObject; // it returns object with debug info if debug is true
```

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f5925b6c95f53653f35ce3f83fcda1ba9b1e04a6/types/adblock-rs/index.d.ts#L27C43-L27C65

diracdeltas:
@antonok-edm

antonok-edm:
interesting, I didn't realize there were 3rd-party type bindings for this repo. Probably the issue should be filed there instead?

I guess the ideal solution would be to migrate the types into this repo... I'm not a TS expert but I will investigate options.

PR #343, opened 2024-03-12
Add `index.d.ts`

antonok-edm:
closes https://github.com/brave/adblock-rust/issues/340

@darahe does this look good to you?

### DIFF
--- a/js/index.d.ts
+++ b/js/index.d.ts
@@ -1,0 +1,122 @@
+/// <reference lib="es2020" />
+
+declare interface MimeType {
+    mime: string;
+}
+
+export interface Resource {
+    name: string;
+    aliases: string[];
+    kind: MimeType | "template";
+    content: string;
+    dependencies?: string[];
+    permission?: number;
+}
+
+export interface FilterListMetadata {
+    homepage?: string;
+    title?: string;
+    expires?: number;
+    redirect?: string;
+}
+
+export interface ParseOptions {
+    format: typeof FilterFormat;
+    rule_types: typeof RuleTypes;
+}
+
+declare interface BlockerResult {
+    matched: boolean,
+    important: boolean,
+    redirect?: string,
+    rewritten_url?: string,
+    exception?: string,
+    filter?: string,
+}
+
+export class Engine {
+    constructor(rules: FilterSet, debug: boolean);
+    addResource(resource: Resource): boolean;
+    check(url: string, source_url: string, request_type: string, debug?: false): boolean;
+    check(url: string, source_url: string, request_type: string, debug: true): BlockerResult;
+    clearTags(): null;
+    deserialize(serialized_handle: ArrayBuffer): null;
+    enableTag(tag: string): null;
+    getResource(name: string): Resource;
+    serializeCompressed(): ArrayBuffer;
+    serializeRaw(): ArrayBuffer;
+    tagExists(tag: string): boolean;
+    useResources(resources: Resource[]): null;
+}
+
+type CbType =
+    'block' |
+    'block-cookies' |
+    'css-display-none' |
+    'ignore-previous-rules' |
+    'make-https';
+
+declare interface CbAction {
+    type: CbType,
+    selector?: string,
+}
+
+type CbResourceType =
+    'document' |
+    'image' |
+    'style-sheet' |
+    'script' |
+    'font' |
+    'raw' |
+    'svg-document' |
+    'media' |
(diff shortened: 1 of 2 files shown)


js/index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+    addFilters(rules: string[], opts?: ParseOptions): FilterListMetadata;
+    intoContentBlocking(): ContentBlockingConversionResult | undefined;
+}
+
+export const FilterFormat: {
+    HOSTS: string;
+    STANDARD: string;
+};
+
+export const RuleTypes: {
```
darahe, line 112: I guess it should be an enum or union type. I do not know which will be compatible with the Rust library. At least it is enum here: https://docs.rs/adblock/latest/adblock/lists/enum.RuleTypes.html

For now, it is an object with all required fields and it can not be used to parse only some types of rules 

https://github.com/brave/adblock-rust/blob/dc2a4fa8bdda4e40f56deef63d4d39c0c925cce2/js/index.d.ts#L103

darahe (approved):
One comment, everything else looks good

js/index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+    redirect?: string,
+    rewritten_url?: string,
+    exception?: string,
+    filter?: string,
+}
+
+export class Engine {
+    constructor(rules: FilterSet, debug: boolean);
+    addResource(resource: Resource): boolean;
+    check(url: string, source_url: string, request_type: string, debug?: false): boolean;
```
darahe, line 40: Also, `request_type` is enum in Rust. So, I think it should be enum here too. 

https://docs.rs/adblock/latest/adblock/request/enum.RequestType.html

js/index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+    check(url: string, source_url: string, request_type: string, debug: true): BlockerResult;
```
darahe, line 41: Nit: we can pass an object (aka named parameters) instead of passing 4 parameters separately, and then we will have a better user experience.

darahe:
@antonok-edm If you need some help to finish this PR you can ping me here

js/index.d.ts
antonok-edm: hmm, after some reading ([1], [2]), my understanding is that `enum`s _shouldn't_ be declared in a `.d.ts` file.

This seems correct to me because it's describing the shape of `(await import('adblock-rs')).RuleTypes`, which basically behaves like a struct with those 3 fields.

Perhaps there is an issue with using `rule_types: typeof RuleTypes` in `ParseOptions` instead? I think it needs to be "field of" rather than `typeof`, but I don't know how to express that.

[1]: https://lukasbehal.com/2017-05-22-enums-in-declaration-files/
[2]: https://www.reddit.com/r/typescript/comments/dp7zd5/enum_or_enumlike_type_in_a_declaration_file/

js/index.d.ts
antonok-edm: technically, although the enum isn't currently used in any public-facing Rust API. Those functions just accept string arguments. I do want to change that but it'll have to be left to a breaking change in the future.

js/index.d.ts
antonok-edm: sounds like that's a breaking change as well, so I'll leave it as-is for now

js/index.d.ts
darahe: Maybe we can deal with union type here: 
```ts
type RequestType =
  | 'beacon'
  | 'csp'
  | 'document'
  | 'dtd'
  | 'fetch'
  | 'font'
  | 'image'
  | 'media'
  | 'object'
  | 'other'
  | 'ping'
  | 'script'
  | 'stylesheet'
  | 'subdocument'
  | 'websocket'
  | 'xlst'
  | 'xmlhttprequest';
  ```

For reference see different implementations from other adblockers:

https://github.com/AdguardTeam/tsurlfilter/blob/master/packages/tsurlfilter/src/request-type.ts#L6-L39

https://github.com/ghostery/adblocker/blob/0eb03a3f412c0764469f7561cf5f734eb4df4c66/packages/adblocker/src/request.ts#L58-L71

js/index.d.ts
darahe: Maybe you can use `keyof typeof RyleTypes` or union type like here:
https://github.com/brave/adblock-rust/pull/343#discussion_r1530839332

PR #352, opened 2024-04-23
Update README.md

KPCOFGS:
Remove `.txt` from LICENSE.txt so when user click on license tag appeared in README, it will redirect users to the actual LICENSE file


antonok-edm (approved):
great catch, thanks!

PR #393, opened 2024-08-30
Disable unicode in network filter regex

Ambyjkl:
According to https://docs.rs/regex/latest/regex/#unicode-can-impact-memory-usage-and-search-speed the unicode flag in regex (which is enabled by default) can have unintended memory consequences. As a result, I observed that disabling this flag resulted in the built network filter regexes from easylist and easyprivacy end up using **a tenth** the memory that it would otherwise.

Unicode characters in urls are converted to ascii using punycode and url component encoding by browsers:
```
https://example-tesT߶.com/tesT߶  ->  https://xn--example-test-1ju.com/tesT%DF%B6
```
Hence, there is no reason to continue having the unicode flag enabled on network filters. All tests continue to pass after disabling it.

Also, as an extra bonus, the `check_matching_equivalent` test, which compares with ublock origin on a real-world dataset, ends up running roughly 5% faster for me, indicating a net performance boost in addition to the memory usage reduction.

Note that we need to use `regex::bytes::Regex` instead of the string version, because the way it's designed, the string version is never allowed to match non-unicode values, so even a single `.` in the regex can cause it to fail to compile since with the unicode flag disabled, `.` can match any byte, including non-printable characters. This is not a cause for concern because we only pass in `String` and `&str` which already don't permit non-utf8 data.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -768,14 +768,15 @@ mod tests {
                                        "script").unwrap();
             assert!(engine.check_network_request(&request).matched);
         }*/
-        {
+        // fails - unicode not supported in network filter
+        /*{
             let engine = Engine::from_rules_debug([r#"/tesT߶/$domain=example.com"#], Default::default());
             let request = Request::new("https://example.com/tesT߶",
                                        "https://example.com",
                                        "script").unwrap();
             assert!(engine.check_network_request(&request).matched);
-        }
-        // fails - punycoded domain
+        }*/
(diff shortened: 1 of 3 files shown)


Quackdoc:
ping @antonok-edm

antonok-edm (approved):
Thanks @Ambyjkl, and sorry for the delay! I double-checked and all URLs should definitely be punycoded at check time, so this is unambiguously a positive change. Great work finding this!

PR #417, opened 2025-01-24
Setup basic perf CI

atuchin-m:
The PR:
1. adds a CI job to estimate perf via [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark/) that fails on +30% regression. The job also pushes the data to GH pages to make graphs.
2 updates easylists*, remove file duplicates and add brave main list in the benchmarks. Also provides a script to do the next updates. 
3. adds a benchmark to measure the time to check the first request (~20ms).

An example (from a forked repo):
image
image


benches/bench_rules.rs
```diff
@@ -84,8 +84,6 @@ fn list_parse(c: &mut Criterion) {
 fn get_blocker(rules: impl IntoIterator<Item=impl AsRef<str>>) -> Blocker {
     let (network_filters, _) = adblock::lists::parse_filters(rules, false, Default::default());
 
-    println!("Got {} network filters", network_filters.len());
```
boocmp, line 87: I'd prefer to comment this line as done in other benches. Or may be leave it as is If it doesn't break anything

atuchin-m: It actually breaks a bench with `--output-format bencher`

data/update-lists.sh
boocmp, line 1: Could it be the nmp command?

src/filters/network.rs
```diff
@@ -3418,7 +3418,7 @@ mod hash_collision_tests {
         let rules = test_utils::rules_from_lists([
             "data/easylist.to/easylist/easylist.txt",
             "data/easylist.to/easylist/easyprivacy.txt",
-        ]);
+        ]).filter(|f| *f != "||www.bred4tula.com^"); // remove known collision
```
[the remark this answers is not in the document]
atuchin-m: done

data/update-lists.sh
atuchin-m: updated

.github/workflows/perf-ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: cargo bench --bench bench_matching rule-match-first-request -- --output-format bencher | tee -a output.txt
+
+      - name: Bench startup speed
+        run: cargo bench --bench bench_rules blocker_new/brave-list -- --output-format bencher | tee -a output.txt
+
+      - name: Bench memory usage
+        run: cargo bench --bench bench_memory -- --output-format bencher | tee -a output.txt
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1.20.4
```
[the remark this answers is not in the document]
atuchin-m: SHA1 can't be used here, the action isn't designed for that:
```
File not found: '/home/runner/work/_actions/benchmark-action/github-action-benchmark/6bae118c112083251560ad8b3a1ff2e43aa23351/dist/src/index.js'
```

https://github.com/brave/adblock-rust/actions/runs/12953281281

I've set `v1.20.4`

data/update-lists.js
```diff
@@ -0,0 +1,53 @@
+const { execSync } = require("child_process");
+const readline = require("readline");
+const fs = require("fs");
+
+const rl = readline.createInterface({
```
boocmp, line 5: wouldn't it be easier to pass the key as a command parameter?

atuchin-m: make sense, done.

boocmp (approved):

.github/workflows/perf-ci.yml
```diff
@@ -0,0 +1,50 @@
...
```
thypon, line 41:
```suggestion
        uses: brave/github-action-benchmark@v1.20.4
```

thypon (changes requested):

.github/workflows/perf-ci.yml
atuchin-m: done

thypon (approved):

data/update-lists.js
```diff
@@ -0,0 +1,44 @@
...
+  "curl -o data/easylist.to/easylistgermany/easylistgermany.txt https://easylist.to/easylistgermany/easylistgermany.txt"
+);
+
+execSync(
+  `curl -o extension.zip -H "BraveServiceKey: ${apiKey}" ` +
+    `https://brave-core-ext.s3.brave.com/release/${extensionId}/extension_${versionNumber}.crx`
+);
+
+try {
+  execSync("unzip extension.zip list.txt");
```
antonok-edm, line 34: I'd either do this part in a temp directory or warn if there are any existing files that could get overwritten

js/Cargo.toml
```diff
@@ -14,3 +14,6 @@ serde =  { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 adblock = { path = "../", features = ["css-validation", "content-blocking", "resource-assembler"] }
 neon = { version = "^0.10.1", default-features = false, features = ["napi-1"] }
+
+[features]
+default-panic-hook = []
```
antonok-edm, line 19: I don't think this does anything?

data/update-lists.js
atuchin-m: Updated to use a temp dir.

js/Cargo.toml
antonok-edm: One of the changes in Rust 1.84 is https://github.com/rust-lang/rust/pull/132577, so this is now a lint warning (and thus an error since we've configured lint warnings as errors). I think it's resolved in newer versions of Neon, but we can leave it here as-is for now.

antonok-edm (approved):

PR #425, opened 2025-02-04
Code structure

boocmp:
The test code has been moved to the test/unit/ folder, replicating the src directory structure. In my opinion, this improves the efficiency of working with the codebase.


tests/unit/optimizer.rs
```diff
@@ -0,0 +1,660 @@
...
+                ""
+            )
+            .unwrap()
+        ))
+    }
+}
+
+/*
+#[cfg(test)]
+mod optimization_tests_union_domain {
```
atuchin-m, line 518: a duplicate? about the same block on line 190.

src/url_parser/parser_regex.rs
```diff
@@ -1,77 +0,0 @@
-use regex::Regex;
-
-pub fn get_hostname_regex(url: &str) -> Option<(usize, (usize, usize))> {
```
atuchin-m, line 3: is it a unused code?

atuchin-m (approved):

src/url_parser/parser_regex.rs
boocmp: Yes

boocmp:
> @boocmp what exactly is the justification for moving all the tests? Codebase churn and personal opinions aside, it's considered idiomatic Rust to keep the unit tests close to the corresponding implementations.

The first reason is to reduce the amount of code in files. It’s difficult to work with a file that has over a thousand lines of code, especially when most of it consists of tests. For example, a task like adding code to the end of a file becomes quite tedious you have to search for where the tests begin instead of simply pressing the End key. The second reason is search. I can configure a search for any substring while ignoring tests by simply excluding the test directory. The third reason is GitHub diffs. It’s much clearer to see where changes are made in the code versus in the tests, since GitHub only shows a few lines around the changes, it’s not always obvious otherwise. I’m not an expert in Rust, but this recommendation goes against everything I’ve seen in other languages and projects throughout my experience.

PR #429, opened 2025-02-11
Silently ignore generic procedural filter

ShivanKaul:
Don't panic when trying to load a generic procedural filter


boocmp (approved):

ShivanKaul:
This case would be caught during parse time, but I still don't think the adblock engine should ever panic. So I think this check is worth making anyway.

PR #434, opened 2025-02-20
Code structure follow up #3.

boocmp:
Bench performance has been improved a little.
`check`, `check_all` functions are simplified.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ seahash = "3"   # seahash 4 introduces a breaking hash algorithm change
 memchr = "2.4"
 base64 = "0.13"
 rmp-serde = "0.15"
-lifeguard = { version = "^ 0.6.1", optional = true }
 cssparser = { version = "0.29", optional = true }
 selectors = { version = "0.24", optional = true }
 serde_json = "1.0"
@@ -89,9 +88,8 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "unsync-regex-caching"]
+default = ["embedded-domain-resolver", "full-regex-handling", "unsync-regex-caching"]
 full-regex-handling = []
-object-pooling = ["lifeguard"] # disables `Send` and `Sync` on `Engine`.
 unsync-regex-caching = [] # disables `Send` and `Sync` on `Engine`.
 regex-debug-info = []
 css-validation = ["cssparser", "selectors"]
--- a/src/request.rs
+++ b/src/request.rs
@@ -87,6 +87,7 @@ pub struct Request {
     pub is_supported: bool,
     pub is_third_party: bool,
     pub url: String,
+    pub url_lower_cased: String,
     pub hostname: String,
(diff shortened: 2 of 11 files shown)


src/request.rs
```diff
@@ -98,15 +100,25 @@ impl Request {
...
-    pub fn get_tokens(&self, token_buffer: &mut Vec<utils::Hash>) {
-        token_buffer.clear();
-        utils::tokenize_pooled(&self.url.to_ascii_lowercase(), token_buffer);
-        // Add zero token as a fallback to wildcard rule bucket
-        token_buffer.push(0);
+    pub fn get_tokens_for_match(
+        &self,
+    ) -> std::iter::Chain<
+        core::iter::Flatten<std::option::IntoIter<&Vec<utils::Hash>>>,
+        std::slice::Iter<'_, utils::Hash>,
```
antonok-edm, line 111: return type of `impl Iterator<Item=&utils::Hash>` would be cleaner here

src/request.rs
```diff
@@ -87,7 +87,9 @@ pub struct Request {
     pub is_supported: bool,
     pub is_third_party: bool,
     pub url: String,
+    pub url_lower_cased: String,
     pub hostname: String,
+    pub request_tokens: Vec<utils::Hash>,
```
antonok-edm, line 92: let's make these new fields `pub(crate)`, they're an internal implementation detail and we don't want them to be set outside of the constructors

src/request.rs
```diff
@@ -90,23 +88,30 @@ pub struct Request {
...
         }
     }
 
-    pub fn get_tokens(&self, token_buffer: &mut Vec<utils::Hash>) {
-        token_buffer.clear();
-        utils::tokenize_pooled(&self.url.to_ascii_lowercase(), token_buffer);
-        // Add zero token as a fallback to wildcard rule bucket
-        token_buffer.push(0);
+    pub fn get_tokens_for_match(&self) -> impl Iterator<Item = &utils::Hash> {
+        self.source_hostname_hashes
```
atuchin-m, line 106: Let's add a small comment that `source_hostname_hashes` should be the first for performance reasons.

Cargo.toml
```diff
@@ -89,9 +88,8 @@ harness = false
 [features]
 # If disabling default features, consider explicitly re-enabling the
 # "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "object-pooling", "unsync-regex-caching"]
```
atuchin-m, line 92: We also need to disable in b-c when the a version is merged:
https://github.com/brave/brave-core/blob/master/components/brave_shields/adblock/rs/Cargo.toml#L15

atuchin-m (approved):

src/request.rs
```diff
@@ -156,10 +161,14 @@ impl Request {
...
         };
 
+        let url_lower_cased = url.to_ascii_lowercase();
+
         Request {
             request_type,
             url: url.to_owned(),
+            url_lower_cased: url_lower_cased.to_owned(),
             hostname: hostname.to_owned(),
+            request_tokens: calculate_tokens(&url_lower_cased),
```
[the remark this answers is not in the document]
boocmp: We create the `Request` only to check it.

PR #439, opened 2025-03-03
Code structure follow up #4

boocmp:
1. The implementation of `NetworkFilterList` has been moved to network_filter_list.rs.
2. The `bitflags` dependency version has been bumped to 2.9.0.
3. The `flatbuffers` dependency has been added.
4. FlatBuffers schema of the network filter list and the corresponding generated file have been added.


Cargo.toml
```diff
@@ -28,7 +28,7 @@ url = "2.5"
 percent-encoding = "2.1"
 once_cell = "1.8"
 regex = "1.5"
-bitflags = "1.3"
+bitflags = { version = "2.9.0", features = ["serde"] }
```
boocmp, line 31: @antonok-edm I’ve bumped the `bitflags` version because `flatbuffers` depends on it, and I want to avoid depending on two different versions. However, this change breaks backward compatibility. There is a crate called [bitflags-serde-legacy](https://crates.io/crates/bitflags-serde-legacy) that provides compatibility for older versions. Should I revert this change and stick with `1.3`? What do you think?

src/flatbuffers/fb_network_filter.fbs
```diff
@@ -0,0 +1,26 @@
+// flatc --rust --gen-object-api -o src/flatbuffers/ src/flatbuffers/fb_network_filter.fbs
+
+namespace fb;
+
+table NetworkFilter {
+  mask: uint32;  // NetworkFilterMask (network.rs)
+
+  // These arrays contain sorted(asc) indecies in the |unique_domains_hashes|
```
atuchin-m, line 8: indecies  => indices
itself => themselves

Maybe something like:
These arrays contain the sorted (ascending) indices of elements in the |unique_domains_hashes| array, rather than the hashes themselves.

Cargo.toml
atuchin-m: IMHO, it makes sense to stick with `1.3` because we're going to break backward compatibility anyway.

atuchin-m (approved):

Cargo.toml
antonok-edm: > There is a crate called [bitflags-serde-legacy](https://crates.io/crates/bitflags-serde-legacy) that provides compatibility for older versions. Should I revert this change and stick with `1.3`? What do you think?

I recall looking into this previously - that crate wasn't even working correctly in my testing, but I wasn't able to find the reason for it.

> IMHO, it makes sense to use the recent `2.9.0` because we're going to break backward compatibility anyway.

Technically there's backward compatibility w/ the API, and then there's backward compatibility with the serialized format. So far we've been careful to keep the serialized format at least continuously usable across minor adblock-rust versions.

We no longer strictly require either for the sake of Brave Browser. I'm not sure how other downstream dependents of adblock-rust treat the serialized formats but I'm not convinced it's worth it to maintain that compatibility going forwards. It only makes sense in a scenario where the underlying lists are not being distributed; other projects likely also have access to the list text.

antonok-edm: btw, if we're doing this, we may as well also upgrade `seahash` which was held back for a similar reason (doesn't have to be in the same PR)

src/blocker.rs
```diff
@@ -2,18 +2,17 @@
...
+use serde::Serialize;
+use std::collections::HashSet;
 use std::ops::DerefMut;
-use std::sync::Arc;
 use thiserror::Error;
 
-use crate::filters::network::{NetworkFilter, NetworkFilterMaskHelper, NetworkMatchable};
-use crate::optimizer;
+use crate::filters::network::{NetworkFilter, NetworkFilterMaskHelper};
+pub(crate) use crate::network_filter_list::NetworkFilterList;
```
antonok-edm, line 11: let's drop the `pub(crate)` here, and change other imports to refer to it directly from `crate::network_filter_list`

src/lib.rs
```diff
@@ -24,6 +24,7 @@ mod data_format;
 mod engine;
 pub mod filters;
 pub mod lists;
+pub mod network_filter_list;
```
antonok-edm, line 27: we can remove `pub` here; nothing inside the module is public

Cargo.toml
boocmp: Ok, then I'll update `bitflags` and `seahash` crates and fix the corresponding tests.

data/rs-ABPFilterParserData.dat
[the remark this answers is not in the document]
boocmp: iOS: I’ve just made the live tests an optional step in the GitHub workflow. At some point, we’ll need to update the CI (or the place where the .dat file for iOS is generated, I’m not sure) to generate the new version. This might also require an update to the browser code.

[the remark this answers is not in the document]
boocmp: @antonok-edm Do you know how and where the `latest.dat` file for iOS is generated? Is it part of the iOS browser, or is it a downloadable component?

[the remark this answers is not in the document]
antonok-edm: > Leftover? Do we use it somewhere?

Right, we can remove `rs-ABPFilterParserData.dat` and related tests safely if we are no longer guaranteeing that older serialized versions can be deserialized. We only need to continue to test that the round trip from list text -> serialized -> `Engine` still works as expected.

> Do you know how and where the latest.dat file for iOS is generated?

iOS switched to using plaintext files quite some time ago so that should be safe to remove as well.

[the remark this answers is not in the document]
antonok-edm: > I’ve just made the live tests an optional step in the GitHub workflow.

IIRC some of the live tests use the list text, we should keep those but anything using `.dat` files can be removed.

Cargo.toml
bridiver: seahash 4.x looks ok, I don't see any new dependencies

antonok-edm:
@atuchin-m my understanding is that on any failure to decode the `.dat` file, the filters will be loaded from the plaintext lists instead. You might want to ask @StephenHeaps for more details on specific error-handling behavior, since this hasn't been part of the "happy-path" thus far.

PR #442, opened 2025-03-13
Bump base64

bridiver:


tests/unit/engine.rs
```diff
@@ -3,6 +3,8 @@ mod tests {
     use super::super::*;
     use crate::lists::FilterFormat;
     use crate::resources::MimeType;
+    use base64::prelude::*;
+    use base64::Engine;
```
antonok-edm, line 8: 1. `base64::Engine` doesn't appear to be used here?
2. instead of importing `base64::prelude::*` in all the files and running into name conflicts with adblock-rust's own structs, let's just selectively import what's actually used? `use base64::prelude::BASE64_STANDARD;` looks like it would be sufficient in all the files.

benches/bench_redirect_performance.rs
```diff
@@ -6,6 +6,9 @@ use adblock::filters::network::{NetworkFilter, NetworkFilterMask, NetworkFilterM
 use adblock::request::Request;
 use adblock::resources::ResourceStorage;
 
+#[cfg(not(feature = "resource-assembler"))]
+use base64::prelude::*;
+
```
antonok-edm, line 11: this could be done inside the existing `cfg` block, next to the other `use` statement there

src/resources/mod.rs
```diff
@@ -20,6 +20,9 @@ pub use resource_storage::{AddResourceError, ResourceStorage, ScriptletResourceE
 use memchr::memrchr as find_char_reverse;
 use serde::{Deserialize, Serialize};
 
+#[cfg(test)]
+use base64::prelude::*;
+
```
antonok-edm, line 25: likewise this could go inside `Resource::simple()`

tests/unit/engine.rs
```diff
@@ -3,6 +3,8 @@ mod tests {
     use super::super::*;
     use crate::lists::FilterFormat;
     use crate::resources::MimeType;
+    use base64::prelude::BASE64_STANDARD;
+    use base64::engine::Engine as _;
```
antonok-edm, line 7: could combine these like `use base64::{prelude::BASE64_STANDARD, engine::Engine as _};`

and I think it'd be good replace all the `prelude::*` imports this way to avoid similar confusion in the future

tests/unit/engine.rs
```diff
@@ -357,7 +360,8 @@ mod tests {
             );
         }
         {
-            let engine = Engine::from_rules_debug(["||example.com^$script"], Default::default());
+            let engine =
+                Engine::from_rules_debug(["||example.com^$script"], Default::default());
```
antonok-edm, line 364: mind reverting the unrelated format changes? I will have to rebase this on our other branch as well

bridiver: won't it fail the format check then? That's why I ran it in the first place

Cargo.lock
```diff
@@ -77,9 +77,15 @@ checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "base64"
-version = "0.21.5"
+version = "0.21.7"
```
bridiver, line 80: omg cargo you are so stupid

antonok-edm (approved):

PR #446, opened 2025-03-28
Flatbuffers impl

boocmp:
1. The `NetworkFilterList` implementation is now based on `FlatBuffers`.
2. The serialization format has been updated to support this change.
3. Serialization benchmarks have been moved to a separate file.
4. Support for adding filters after `NetworkFilterList` creation has been removed. Since rebuilding the entire `FlatBuffer` is costly, this functionality may be reintroduced in the future if there is a clear need for it.


src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn new() -> Self {
+        Self {
+            builder: flatbuffers::FlatBufferBuilder::new(),
+            filters: vec![],
+            unique_domains: vec![],
+            unique_domains_map: HashMap::new(),
+        }
+    }
+
+    fn get_or_insert(&mut self, h: &Hash) -> u16 {
```
atuchin-m, line 38: nit: `get_or_insert_unique_domain`

.github/workflows/perf-ci.yml
```diff
@@ -37,6 +37,18 @@ jobs:
       - name: Bench memory usage
         run: cargo bench --bench bench_memory -- --output-format bencher | tee -a output.txt
 
+      - name: Bench network filter matching (flat)
```
atuchin-m, line 40: let's remove this in favor of enabling `flatbuffers-storage` by default (or in the tests above).

src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+                .filter
+                .iter()
+                .map(|s| self.builder.create_string(s))
+                .collect();
+            Some(self.builder.create_vector(&offsets))
+        } else {
+            None
+        };
+
+        let raw_line = network_filter
```
atuchin-m, line 89: do we create this only in debug mode?

src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn finish(&mut self) -> Vec<u8> {
+        let filters = self.builder.create_vector(&self.filters);
+
+        let unique_domains = self.builder.create_vector(&self.unique_domains);
+
+        let storage = fb::NetworkFilterList::create(
+            &mut self.builder,
+            &&fb::NetworkFilterListArgs {
+                network_filters: Some(filters),
+                unique_domains_hashes: Some(unique_domains),
```
atuchin-m, line 121: We don't convert string to hashes here, so let's replace `unique_domains_hashes` to `unique_domains` or vice versa

src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+        self.len
+    }
+}
+
+pub struct FlatNetworkFilter<'a> {
+    key: u64,
+    owner: &'a FlatNetworkFilterList,
+    fb_filter: &'a fb::NetworkFilter<'a>,
+
+    pub mask: NetworkFilterMask,
```
atuchin-m, line 187: it's not clear why `mask` is pub here, but the other fields are not.

src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+impl<'a> FlatPatterns<'a> {
+    #[inline(always)]
+    pub fn new(
+        patterns: Option<flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<&'a str>>>,
+    ) -> Self {
+        Self { patterns }
+    }
+
+    #[inline(always)]
+    pub fn iter(&self) -> FlatPatternsIterator {
```
atuchin-m, line 143: It's not absolutely clear why we implement iterator traits instead using the structure as-is.
If we do it to use a one `check_pattern()` for both "old" and flat impl, let's make a comment to inline this after "old" impl is gone.

atuchin-m (approved):
LGTM with a few nits

src/flat_network_filter_list.rs
```diff
@@ -0,0 +1,199 @@
...
+        request: &Request,
+        active_tags: &HashSet<String>,
+        regex_manager: &mut RegexManager,
+    ) -> Option<CheckResult> {
+        if self.filter_map.is_empty() {
+            return None;
+        }
+
+        let filters_list =
+            unsafe { fb::root_as_network_filter_list_unchecked(&self.flatbuffer_memory) };
```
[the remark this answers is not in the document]
atuchin-m: @boocmp use `root_as_network_filter_list()` + `.expect()` to remove `unsafe`

src/filters/fb_network.rs
boocmp: yes only debug

src/filters/fb_network.rs
boocmp: because I removed direct dependency from any storages in the check functions. I'm not planning to inline it

.github/workflows/perf-ci.yml
boocmp: in the next pr

src/filters/fb_network.rs
boocmp: because it is used outside, in the `FlatNetworkFilterList::check`

atuchin-m: should it be `pub(crate)`?

boocmp: it could. done

src/flat_network_filter_list.rs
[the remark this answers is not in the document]
boocmp: it runs a verification process, it'll degrade performance.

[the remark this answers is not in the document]
atuchin-m: Until we build the flatbuffer just before using `root_as_network_filter_list_unchecked` should be fine in terms of security.

src/filters/fb_network.rs
```diff
@@ -0,0 +1,307 @@
...
+    #[inline(always)]
+    pub fn include_domains(&self) -> Option<&[u16]> {
+        self.fb_filter.opt_domains().map(|data| {
+            let bytes = data.bytes();
+            unsafe {
+                std::slice::from_raw_parts(
+                    bytes.as_ptr() as *const u16,
+                    bytes.len() / std::mem::size_of::<u16>(),
+                )
+            }
```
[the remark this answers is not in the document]
boocmp: assert added

src/filters/fb_network.rs
```diff
@@ -0,0 +1,313 @@
...
+            o.dedup();
+            self.builder.create_vector(&o)
+        });
+
+        let opt_not_domains = network_filter.opt_not_domains.as_ref().map(|v| {
+            let mut o: Vec<u16> = v
+                .iter()
+                .map(|x| self.get_or_insert_unique_domain_hash(x))
+                .collect();
+            o.sort_unstable();
```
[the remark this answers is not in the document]
boocmp: Here, we're constructing an index of unique hashes to eliminate duplicates and save memory. This index is implemented as a sorted array to enable binary search.

src/flat_network_filter_list.rs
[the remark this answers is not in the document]
boocmp: added this check on deserialization

src/engine_serializer.rs
```diff
@@ -0,0 +1,54 @@
...
+        DeserializeError::DataFormatError(value)
+    }
+}
+
+pub trait EngineSerializer {
+    fn serialize(&self) -> Result<Vec<u8>, SerializeError>;
+    fn deserialize(&mut self, serialized: &[u8]) -> Result<(), DeserializeError>;
+}
+
+impl EngineSerializer for Engine {
```
antonok-edm, line 30: could we leave these directly on `impl Engine`? I see no reason to move it into a separate trait; just makes the API slightly more confusing for users

src/filters/fb_network.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl<'a> FlatNetworkFilter<'a> {
+    #[inline(always)]
+    pub fn new(
+        filter: &'a fb::NetworkFilter<'a>,
+        index: u32,
+        owner: &'a NetworkFilterList,
+    ) -> Self {
+        let list_address: *const NetworkFilterList = owner as *const NetworkFilterList;
```
antonok-edm, line 202: pointer casting often indicates some problematic design patterns in Rust. I think we can remove it in this instance?
- this `owner` field can increase memory usage since it's an additional address-sized field on each filter, even though they'd all have the same value for all filters in a list
- I only see it used in `NetworkMatchable` to make it conform to the existing API. I'm not opposed to that API changing if needed.

src/filters/fb_network.rs
```diff
@@ -0,0 +1,314 @@
...
+            &mut self.builder,
+            &fb::NetworkFilterArgs {
+                mask: network_filter.mask.bits(),
+                patterns: patterns,
+                modifier_option: modifier_option,
+                opt_domains: opt_domains,
+                opt_not_domains: opt_not_domains,
+                hostname: hostname,
+                tag: tag,
+                raw_line: raw_line,
```
antonok-edm, line 110: these can all use shorthand field initialization

src/network_filter_list.rs
```diff
@@ -1,37 +1,94 @@
...
+}
+
+impl Default for NetworkFilterList {
+    fn default() -> Self {
+        Self {
+            flatbuffer_memory: Default::default(),
+            filter_map: Default::default(),
+            unique_domains_hashes_map: Default::default(),
+        }
+    }
```
antonok-edm, line 62: `#[derive(Serialize, Deserialize, Default)]` should automatically produce this `Default` implementation already

src/network_filter_list.rs
```diff
@@ -1,37 +1,94 @@
...
+
+impl NetworkFilterMaskHelper for CheckResult {
+    #[inline]
+    fn has_flag(&self, v: NetworkFilterMask) -> bool {
+        self.filter_mask.contains(v)
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct NetworkFilterList {
```
antonok-edm, line 49: can we keep the original `pub(crate)` visibility?

tests/unit/blocker.rs
```diff
@@ -1289,135 +1289,20 @@ mod blocker_tests {
...
-            blocker.add_filter(filter.clone()).unwrap();
-            assert!(
-                blocker.filter_exists(&filter),
-                "Expected filter to be inserted"
-            );
-            let added = blocker.add_filter(filter);
-            assert!(added.is_err(), "Expected repeated insertion to fail");
-            assert_eq!(
-                added.err().unwrap(),
-                BlockerError::FilterExists,
```
antonok-edm, line 1326: `BadFilterAddUnsupported` and `FilterExists` error variants can be removed now

src/filters/fb_network.rs
boocmp: `FlatNetworkFilter` is a wrapper over the `fb::NetworkFilter`.  We only create the instance of it when found the the `fb::NetworkFilter` in the `NetworkFilterList` so it does't increase the memory usage because it temporary object on the stack. For now, regex manager is designed to use unique key to find corresponding regex, previously it was an address of the `NetworkFilter`, but now we don't have unique address for every filter because they stored in flatbuffer. To create unique key we use the list address combined with the filter's index. I agree it looks ugly but I don't know how to do it better.

atuchin-m: Let's understand why we need this unique `key`.
If it's only for regex manager, let's:
1. try to move key calculation here
2. add TODO to make keys unique only in scope of regex manager. I believe once that in one of the next steps we move from multiple flatbuffers to a just one that ables to contain multiple lists. That way we get 1:1 proportion (regex manager: flatbuffer).

atuchin-m: Also, this could be done in follow up: this PR is already waiting for a long time.

src/data_format/v0.rs
```diff
@@ -369,24 +374,19 @@ impl From<NetworkFilterV0DeserializeFmt> for NetworkFilter {
...
-    pub filter_map: HashMap<crate::utils::Hash, Vec<NetworkFilterV0DeserializeFmt>>,
+    pub flatbuffer_memory: Vec<u8>,
+    pub filter_map: HashMap<crate::utils::Hash, Vec<u32>>,
+    pub unique_domains_hashes_map: HashMap<crate::utils::Hash, u16>,
 }
 
 impl From<NetworkFilterListV0DeserializeFmt> for NetworkFilterList {
     fn from(v: NetworkFilterListV0DeserializeFmt) -> Self {
+        let _ = fb::root_as_network_filter_list(&v.flatbuffer_memory)
+            .expect("Flatbuffer is not corrupted");
```
atuchin-m, line 385: Is it result in `panic()` if the format doesn't match?
We definitely don't want to crash the browser process here.

boocmp: Implemented TryFrom instead

src/data_format/mod.rs
```diff
@@ -9,9 +9,10 @@ mod v0;
 
 pub(crate) mod utils;
 
-use crate::blocker::Blocker;
 use crate::cosmetic_filter_cache::CosmeticFilterCache;
 
+use crate::blocker::Blocker;
+
```
antonok-edm, line 15: nit but I don't understand the reason to move `Blocker` into a separate paragraph by itself (similar change also in `src/data_format/v0.rs`)

antonok-edm (approved):
the other thing left unaddressed is https://github.com/brave/adblock-rust/pull/446#discussion_r2049462034, still not sure if it's required

src/filters/fb_network.rs
antonok-edm: `1 regex manager: 1 flatbuffer` definitely sounds like the right way to do it, but agreed it can be followup if needed

src/blocker.rs
```diff
@@ -377,13 +389,13 @@ impl Blocker {
             return None;
         }
 
-        let mut disabled_directives: HashSet<&str> = HashSet::new();
-        let mut enabled_directives: HashSet<&str> = HashSet::new();
+        let mut disabled_directives: HashSet<String> = HashSet::new();
+        let mut enabled_directives: HashSet<String> = HashSet::new();
```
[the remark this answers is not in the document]
boocmp: Because the return type of `check` and `check_all` was changed to use the `CheckResult` struct, it's a temporary object so it doesn't live long enough to use `&str`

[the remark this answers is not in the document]
boocmp: returned `&str` back

src/network_filter_list.rs
boocmp: Oh,,god, I forgot to add `Default` it the derive list... I'll fix it in one of the follow ups

PR #464, opened 2025-05-31
Move filter_map to flatbuffers

atuchin-m:
The PR encodes `filter_map` (the largest sedre HashMap we store) into flatbuffer as a pair of vector.
The PR:
* saves -5.5MB memory by replacing the hash map to more memory efficient storage.
* increases lookup time now O(log N) instead of O(1), but it's not a hot path. According to profiler measurements and the perf tests this has no effect to the metrics. 
* moves all unsafe code  to a dedicated file and commented. 
* replaces `Vec<u8>` to `VerifiedFlatFilterListMemory` to wrap the flatbuffer memory in more a safe manner.
* adds a new test `bench_rule_matching_browserlike` to count blocked/passed requests on the browser default list to get a better coverage.

### DIFF
--- a/src/filters/flat_filter_map.rs
+++ b/src/filters/flat_filter_map.rs
@@ -24,7 +24,7 @@ where
     type Item = (usize, <V as Follow<'a>>::Inner);
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.current_index < self.values.len() {
+        if self.current_index < self.indexes.len() {
             if self.indexes[self.current_index] != self.key {
                 return None;
             }
@@ -39,8 +39,15 @@ where
 }
 
 impl<'a, I: PartialOrd + Copy, V> FlatFilterMap<'a, I, V> {
+    // Construct FlatFilterMap from two vectors:
+    // - index: sorted array of keys
+    // - values: array of values, same length as index
     pub fn new(index: &'a [I], values: Vector<'a, ForwardsUOffset<V>>) -> Self {
-        assert!(index.len() == values.len());
+        // Sanity check the size are equal. Note: next() will handle |values| correctly.
+        debug_assert!(index.len() == values.len());
+
+        debug_assert!(index.is_sorted());
+
         Self { index, values }
     }
 


src/filters/unsafe_tools.rs
atuchin-m, line 1: @thypon @kdenhartog 
Could you take a look at this?
The PR doesn't introduce more `unsafe`, but it reorganize `unsafe` usage.
We use unsafe in:
1.  `unsafe { fb::root_as_network_filter_list_unchecked(&self.data()) }` to skip a CPU-intensive flatbuffer verification when the buffer has been just created by our code (= it isn't external data)
2. `unsafe { std::slice::from_raw_parts(...)}` to convert flatbuffer::vector to slice. It's unsafe, because the buffer MUST be properly aligned. To enforce this `assert(..)` were added.

tests/unit/engine.rs
```diff
@@ -172,76 +172,24 @@ mod tests {
         });
     }
 
-    #[test]
-    fn deserialization_backwards_compatible_plain() {
```
atuchin-m, line 176: Actually, we don't need these tests. The other `deserialization_*` tests now verify that we produce the same .dat as before and we can deserialize it.
If the format is changed, we don't want preserve ability to read the one one. Instead, we should update `ADBLOCK_RUST_DAT_VERSION`.

IMHO, better to focus on speeding up building the ruleset from .txt files rather than provide backward-compatibility in some cases. We're still rebuilding the ruleset from scathes every list update and it happens very often.

tests/unit/engine.rs
```diff
@@ -268,6 +216,18 @@ mod tests {
         engine.deserialize(&serialized).unwrap();
     }
 
+    #[test]
+    fn deserialization_brave_list() {
+        let rules = rules_from_lists(&["data/brave/brave-main-list.txt"]);
+        let mut engine = Engine::from_rules_parametrised(rules, Default::default(), false, true);
+        let serialized = engine.serialize().unwrap();
+
+        // TODO: investigate why the hash is different each time and reenable
```
atuchin-m: it's a serde-level issue, flatbuffer memory is stable.

src/filters/fb_network.rs
```diff
@@ -115,22 +116,44 @@ impl<'a> FlatNetworkFiltersListBuilder<'a> {
...
+                filter_map_index: Some(filter_map_index),
+                filter_map_values: Some(filter_map_values),
                 unique_domains_hashes: Some(unique_domains_hashes),
             },
         );
         self.builder.finish(storage, None);
 
-        let binary = Vec::from(self.builder.finished_data());
-        binary
+        // TODO: consider using builder.collapse() to avoid reallocating memory.
```
atuchin-m, line 154: Currently, the builder holds 3MB before the begging data for an unclear reason.
If we start to reused now, we will get +3MB.

src/data_format/mod.rs
```diff
@@ -33,8 +34,8 @@ pub enum DeserializationError {
     RmpSerdeError(rmp_serde::decode::Error),
     UnsupportedFormatVersion(u8),
     NoHeaderFound,
-    InvalidFlatBuffer(flatbuffers::InvalidFlatbuffer),
-    FlatbufferSemanticError,
+    LegacyFormatNoLongerSupported, // not used, for backward compatibility
+    FlatBufferParsingError(crate::network_filter_list::FlatBufferParsingError),
```
boocmp: nit: It is already imported, you can use a short name.

atuchin-m: done

tests/unit/engine.rs
atuchin-m: I've found and fixed the reason.

boocmp (approved):

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,72 @@
+//! Unsafe utility functions for working with flatbuffers and other low-level operations.
+
+use crate::filters::fb_network::flat::fb;
+
+// Minimum alignment for the beginning of the flatbuffer data.
+// Should be 4 until we support armv7 and x86_32.
+const MIN_ALIGNMENT: usize = 4;
```
antonok-edm: comment is wrong (as discussed)

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,72 @@
...
+/// # Safety
+/// This function uses unsafe code to convert flatbuffer vector bytes to a slice.
+/// It asserts the vector data is properly aligned and sized.
+#[inline(always)]
+pub fn fb_vector_to_slice<'a, T>(vector: flatbuffers::Vector<'a, T>) -> &'a [T] {
+    let bytes = vector.bytes();
+
+    const fn static_assert_alignment<T>() {
+        // We can't use T with the size more than MIN_ALIGNMENT.
+        // Since the begging of flatbuffer data is aligned to that size,
```
antonok-edm: nit: `begging` -> `beginning`

src/filters/mod.rs
```diff
@@ -5,4 +5,6 @@ mod network_matchers;
 
 pub mod cosmetic;
 pub mod fb_network;
+pub mod flat_filter_map;
 pub mod network;
+pub mod unsafe_tools;
```
antonok-edm: both can be `pub(crate)`

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,72 @@
+//! Unsafe utility functions for working with flatbuffers and other low-level operations.
+
+use crate::filters::fb_network::flat::fb;
+
+// Minimum alignment for the beginning of the flatbuffer data.
+// Should be 4 until we support armv7 and x86_32.
```
atuchin-m:
```suggestion
// Should be 4 while we support armv7 and x86_32.
```

antonok-edm (approved):

src/filters/mod.rs
atuchin-m: fixed

src/filters/unsafe_tools.rs
atuchin-m: thanks, fixed.

src/filters/unsafe_tools.rs
atuchin-m: thanks, fixed.

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,50 @@
...
+pub fn fb_vector_to_slice<'a, T>(vector: flatbuffers::Vector<'a, T>) -> &'a [T] {
+    let bytes = vector.bytes();
+    assert!(bytes.len() % std::mem::size_of::<T>() == 0);
+    assert!(bytes.as_ptr() as usize % std::mem::align_of::<T>() == 0);
+    unsafe {
+        std::slice::from_raw_parts(
+            bytes.as_ptr() as *const T,
+            bytes.len() / std::mem::size_of::<T>(),
+        )
+    }
```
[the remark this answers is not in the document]
kdenhartog: not concerned with this it's a refactor of this: https://github.com/brave/adblock-rust/pull/464/files#diff-6b8e5b6785874b51889641d444ce31c74d68819555fb2dc1d221bdbaff2e0bd9L223-L247

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,49 @@
...
```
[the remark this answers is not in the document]
kdenhartog: not concerned with this it's a refactor of this: https://github.com/brave/adblock-rust/pull/464/files#diff-6b8e5b6785874b51889641d444ce31c74d68819555fb2dc1d221bdbaff2e0bd9L223-L247

src/filters/unsafe_tools.rs
```diff
@@ -0,0 +1,72 @@
...
+    const fn static_assert_alignment<T>() {
+        // We can't use T with the size more than MIN_ALIGNMENT.
+        // Since the beginning of flatbuffer data is aligned to that size,
+        // the alignment of the data must be a divisor of MIN_ALIGNMENT.
+        assert!(MIN_ALIGNMENT % std::mem::size_of::<T>() == 0);
+    }
+    let _ = static_assert_alignment::<T>;
+
+    assert!(bytes.len() % std::mem::size_of::<T>() == 0);
+    assert!(bytes.as_ptr() as usize % std::mem::align_of::<T>() == 0);
```
kdenhartog, line 26: I see we've previously paniced on this before with the assert macro, but I'm wondering if it would be better to match here instead of panic now that we're supporting arbitrary memory sizes rather than [fixed to u16](https://github.com/brave/adblock-rust/pull/464/files#diff-6b8e5b6785874b51889641d444ce31c74d68819555fb2dc1d221bdbaff2e0bd9L226) byte arrays.

Also, do we need to support arbitrary byte array sizes here or can we keep it [fixed with SIZE_U16 consts](https://docs.rs/flatbuffers/25.2.10/flatbuffers/constant.SIZE_U16.html)?

src/flatbuffers/fb_network_filter_generated.rs
```diff
@@ -465,29 +466,51 @@ pub mod fb {
...
-                network_filters,
+                filter_map_index,
+                filter_map_values,
                 unique_domains_hashes,
             }
         }
 
         #[inline]
-        pub fn network_filters(
+        pub fn filter_map_index(&self) -> flatbuffers::Vector<'a, u32> {
```
kdenhartog, line 499: Is it worth returning an option here so we can bubble the error up and handle it by the calling function rather than panicing with the unwrap in here?

It seems like this is probably fine since the Flatbuffere is being verified, but wanted to check.

src/filters/flat_filter_map.rs
```diff
@@ -0,0 +1,64 @@
...
+            Some((index, filter))
+        } else {
+            None
+        }
+    }
+}
+
+impl<'a, I: PartialOrd + Copy, V> FlatFilterMap<'a, I, V> {
+    pub fn new(index: &'a [I], values: Vector<'a, ForwardsUOffset<V>>) -> Self {
+        assert!(index.len() == values.len());
```
kdenhartog: Should we return an error here rather than panicking if this doesn't match so that it can be handled by the caller?

atuchin-m: It's an invariant and we should never get the situation when they are not equal.
Although, `self.values.get()` already handles `values` overflow. So it makes sense to replace it to debug_assert.

src/filters/unsafe_tools.rs
atuchin-m: It's a code invariant that should never be broken. If one of the asserts is broken it's UB. Chromium C++ style advices to check invariants via CHECK macros (=runtime assert!): https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md#check_dcheck_and-notreached.

There is no good way to recovery from the situation when it's broekn and we need a crash dump to address this if it happens sometimes.

PR #467, opened 2025-06-02
Add ADBLOCK_RUST_DAT_VERSION

atuchin-m:
The PR adds a single point for updating DAT version.


src/flatbuffers/fb_network_filter.fbs
```diff
@@ -1,3 +1,7 @@
+// IMPORTANT NOTE: In case of breaking changes increase ADBLOCK_RUST_DAT_VERSION
+// This helps to speed up browser startup process after the update.
```
antonok-edm, line 2: (more importantly, it makes sure we don't misinterpret the data in case the format is similar enough to still parse)

tests/unit/engine.rs
```diff
@@ -174,13 +174,23 @@ mod tests {
...
     #[test]
     fn deserialization_backwards_compatible_plain() {
-        // deserialization_generate_simple();
-        // assert!(false);
-        // converted from the legacy compressed format
-        let serialized = std::fs::read("tests/unit/data/backwards_compatible_plain.dat")
-            .expect("Failed to read backwards_compatible_plain.dat");
+        const FILE: &str = "tests/unit/data/backwards_compatible_plain.dat";
+        let serialized =
+            std::fs::read(FILE).expect("Failed to read test backwards_compatible_plain.dat");
```
antonok-edm, line 179: fwiw, `unwrap` is totally fine for test code like this

tests/unit/engine.rs
```diff
@@ -174,13 +174,23 @@ mod tests {
...
+        match deserialized_engine.deserialize(&serialized) {
+            Err(e) => {
+                let engine = Engine::from_rules(&["ad-banner"], Default::default());
+                let failed_file = format!("{}.expected", FILE);
+                std::fs::write(&failed_file, engine.serialize().unwrap()).unwrap();
+
+                println!("Looks the current format is not compatible with the previous");
+                println!("Consider updating ADBLOCK_RUST_DAT_VERSION");
+                println!("The expected content is written to: {}", failed_file);
+                assert!(false, "Deserialization error: {:?}", e);
```
antonok-edm: the formatting will be slightly different on failures here; better if the `println` messages can be combined into the `assert` message

src/data_format/mod.rs
```diff
@@ -96,22 +93,24 @@ impl DeserializeFormat {
...
-            match version {
-                0 => Ok(Self::V0(v0::DeserializeFormat::deserialize(serialized)?)),
-                v => Err(DeserializationError::UnsupportedFormatVersion(v)),
-            }
-        } else if serialized.starts_with(&FLATE2_GZ_HEADER_BYTES) {
-            Err(DeserializationError::LegacyFormatNoLongerSupported)
-        } else {
-            Err(DeserializationError::NoHeaderFound)
-        }
+        Ok(Self::V0(v0::DeserializeFormat::deserialize(serialized)?))
```
antonok-edm: the `DeserializeFormat` and `SerializeFormat` enums should be changed into structs instead. Their only purpose is to wrap/unwrap the header data now, no need to select between different versions.

tests/unit/engine.rs
atuchin-m: I know.
The idea is to write the expected file to easily update expectations.
i.e.:
```
cp tests/unit/data/backwards_compatible_tags.dat.failed tests/unit/data/backwards_compatible_tags.dat 
```

tests/unit/engine.rs
atuchin-m: done

src/data_format/mod.rs
atuchin-m: Done.

src/flatbuffers/fb_network_filter.fbs
atuchin-m: actually, it shouldn't be a case: we verify flatbuffer before usage (i.e. it could be corrupted because it was written only partially)

antonok-edm: I don't know what the odds of arbitrary data parsing successfully into a given flatbuffer specification are. But regardless of that, there's a real possibility of a semantic change of an existing field without any other structural modifications; that would still break backward compatibility.

atuchin-m: That's true.
The chance of accepting a old version as the new one is low, but exists.
Increasing the version is the cleanest and safest approach.

PR #468, opened 2025-06-02
Update some deps to address sec alerts

atuchin-m:
`cargo tree` output: 
[cargo_tree.txt](https://github.com/user-attachments/files/20558231/cargo_tree.txt)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,14 +31,14 @@ regex = "1.5"
 bitflags = { version = "2.9.0", features = ["serde"] }
 itertools = "0.13"
 idna = "1.0.3"
-serde =  { version = "1.0", features = ["derive", "rc"] }
+serde =  { version = "1.0.219", features = ["derive", "rc"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -31,14 +31,14 @@ regex = "1.5"
...
-serde =  { version = "1.0", features = ["derive", "rc"] }
+serde =  { version = "1.0.219", features = ["derive", "rc"] }
 seahash = "4.1.0"
 memchr = "2.4"
 base64 = "0.22"
 rmp-serde = "0.15"
 cssparser = { version = "0.29", optional = true }
 selectors = { version = "0.24", optional = true }
-serde_json = "1.0"
+serde_json = "1.0.140"
```
atuchin-m, line 41: it's the same version used in chromium 137:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/third_party/rust/chromium_crates_io/vendor/serde_json-v1/

PR #476, opened 2025-06-05
Skip slow tests in debug mode

atuchin-m:
The PR:
* skips the slowest tests in debug mode. We also run tests in release, it should be enough.
* moves `fuzz`, `audit` and `asan` steps to a dedicated workflow to speed up the builds.

Slow tests that were additionally disable in debug:
```
────────────
     Summary [ 382.888s] 227 tests run: 227 passed (2 slow), 2 skipped
        SLOW [ 197.113s] adblock::ublock-coverage check_matching_equivalent
        SLOW [ 382.006s] adblock::ublock-coverage check_matching_hostnames
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: CI
 
 on:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   ci:
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -2,6 +2,7 @@ name: Sanity
 
 on:
   pull_request:
+  workflow_dispatch:
 
 jobs:
   sanity:


tests/matching.rs
```diff
@@ -150,7 +150,6 @@ fn check_engine_matching() {
 }
 
 #[test]
-#[cfg(not(debug_assertions))] // This test is too slow to run in debug mode
```
boocmp: What's wrong with this?

boocmp (approved):

tests/matching.rs
atuchin-m: The idea was to continue building this code in debug.
Well, we could stick to `#[cfg(not(debug_assertions))] ` if you like preferrer this option.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: CI
 
-on: [push]
+on:
+  pull_request:
```
antonok-edm, line 4: imo, `on: [push]` is still nice for testing branches before a PR is ready

antonok-edm (approved):

.github/workflows/ci.yml
atuchin-m: Why don't just create a draft?
Now it generates a ton of notifications of every branch push.

antonok-edm: maybe we can add a `workflow_dispatch` also, so it can be triggered manually if needed?

atuchin-m: Done.

PR #477, opened 2025-06-05
Cargo clippy fixes

atuchin-m:
The PR generated by the following sequence:
* Applies `cargo clippy --fix`
* Reverts 2 patches that break compilations or tests
* Add the rest unfixed of clippy rules in Cargo.toml with ` = "allow"`
* `cargo fmt`


src/flatbuffers/fb_network_filter_generated.rs
```diff
@@ -255,7 +255,7 @@ pub mod fb {
         pub tag: Option<flatbuffers::WIPOffset<&'a str>>,
         pub raw_line: Option<flatbuffers::WIPOffset<&'a str>>,
     }
-    impl<'a> Default for NetworkFilterArgs<'a> {
+    impl Default for NetworkFilterArgs<'_> {
```
antonok-edm: we probably don't want to apply clippy to `_generated.rs` files?

atuchin-m: ops, missed that part. Fixed.

src/flatbuffers/fb_network_filter_generated.rs
```diff
@@ -1,5 +1,6 @@
 // automatically generated by the FlatBuffers compiler, do not modify
 
+#![allow(clippy::all)]
```
antonok-edm: I think we can do it as an outer attribute in `src/filters/fb_network.rs` like this:
```
#[allow(clippy::all)]
pub mod flat;
```
(which probably also ought to be `pub(crate)`)

atuchin-m: A nice idea! Done.

src/flatbuffers/fb_network_filter.fbs
```diff
@@ -2,8 +2,8 @@
 // This helps to speed up browser startup process after the update.
 //
 // To build *_generated.rs run:
-// flatc --rust --gen-object-api -o src/flatbuffers/ src/flatbuffers/fb_network_filter.fbs
-
+// 1. flatc --rust --gen-object-api -o src/flatbuffers/ src/flatbuffers/fb_network_filter.fbs
+// 2. cargo fmt
```
antonok-edm, line 6: should we also use a `#[rustfmt::skip]`?

atuchin-m: Sadly, we can't.
cargo fmt format all files regardless they are imported or not.
`rustfmt.toml / ignore` also isn't an options, because it's an unstable feature (nightly only)

PR #478, opened 2025-06-05
Add safe alignment for flatbuffer

atuchin-m:
Actually, we don't have strong guaranties that `vec<u8>` will be aligned to 4 bytes on 32-bit architecture. It's an allocator-specific behavior despite I doubt that it's possible in production.

The PR adds a backup path if the vector isn't aligned.

Founded via https://github.com/rust-lang/miri.
`cargo +nightly miri test blocker::unit_tests::blocker_tests::get_csp_directives -- --nocapture`

P.S. The issue was presented in the original flatbuffer branch.

### DIFF
--- a/src/filters/unsafe_tools.rs
+++ b/src/filters/unsafe_tools.rs
@@ -68,24 +68,22 @@ impl VerifiedFlatFilterListMemory {
     // Properly align the buffer to MIN_ALIGNMENT bytes.
     pub(crate) fn from_vec(mut vec: Vec<u8>) -> Self {
         let shift = vec.as_ptr() as usize % MIN_ALIGNMENT;
-        if shift == 0 {
-            // Fast path, the buffer is already aligned.
-            Self {
-                raw_data: vec,
-                start: 0,
-            }
+        let start = if shift == 0 {
+            0
         } else {
             vec.reserve(vec.len() + MIN_ALIGNMENT); // vec.as_ptr() is changed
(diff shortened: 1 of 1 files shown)


src/filters/unsafe_tools.rs
```diff
@@ -42,23 +42,49 @@ pub fn fb_vector_to_slice<'a, T>(vector: flatbuffers::Vector<'a, T>) -> &'a [T]
...
+        } else {
+            vec.reserve(vec.len() + MIN_ALIGNMENT); // vec.as_ptr() is changed
+            let shift = vec.as_ptr() as usize % MIN_ALIGNMENT;
+            let padding = MIN_ALIGNMENT - shift;
+            vec.splice(0..0, vec![0u8; padding]);
+            let memory = Self {
+                raw_data: vec,
+                start: padding,
+            };
+            assert!(memory.data().as_ptr() as usize % MIN_ALIGNMENT == 0);
```
antonok-edm: does it make sense to put the assertion outside, unconditionally? i.e.

```rust
let start = if shift == 0 {
    0
} else {
    // ...fix the vec...
    padding
};
let memory = Self {
    raw_data: vec,
    start,
};
assert!(memory.data().as_ptr() as usize % MIN_ALIGNMENT == 0);
memory
```

atuchin-m: nice, done.

PR #480, opened 2025-06-08
Fix some trivial clippy warnings

atuchin-m:


src/resources/resource_assembler.rs
```diff
@@ -36,10 +36,10 @@ enum ResourceAliasField {
 }
 
 impl ResourceAliasField {
-    fn to_vec(self) -> Vec<String> {
+    fn to_vec(&self) -> Vec<String> {
         match self {
-            Self::SingleString(s) => vec![s],
-            Self::ListOfStrings(l) => l,
+            Self::SingleString(s) => vec![s.clone()],
+            Self::ListOfStrings(l) => l.clone(),
```
antonok-edm: I'd rather not change the code to add cloning if it's not necessary. Can we fix it by renaming as `into_vec` instead?

atuchin-m: It though we have to do self cloning before (and cloning particular fields is better than it).
Let me check

atuchin-m: Done

antonok-edm (approved):

PR #484, opened 2025-06-11
Fix the rest clippy issues

atuchin-m:
The PR cleans up the most of clippy issues.

### DIFF
--- a/src/content_blocking.rs
+++ b/src/content_blocking.rs
@@ -253,6 +253,7 @@ fn non_empty(v: Vec<String>) -> Option<Vec<String>> {
 /// into multiple rules.
 ///
 /// The contained rules can be accessed using `IntoIterator`.
+#[allow(clippy::large_enum_variant)]
 pub enum CbRuleEquivalent {
     /// In most successful cases, an ABP rule can be converted into a single content blocking rule.
     SingleRule(CbRule),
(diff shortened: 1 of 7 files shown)


src/content_blocking.rs
```diff
@@ -260,7 +260,7 @@ pub enum CbRuleEquivalent {
     /// `Document`, and no load type is specified, then the rule should be split into two content
     /// blocking rules: the first has all original resource types except `Document`, and the second
     /// only specifies `Document` with a third-party load type.
-    SplitDocument(CbRule, CbRule),
+    SplitDocument(CbRule, Box<CbRule>),
```
atuchin-m:
```
error: large size difference between variants
   --> src/content_blocking.rs:256:1
    |
256 | / pub enum CbRuleEquivalent {
257 | |     /// In most successful cases, an ABP rule can be converted into a single content blocking rule.
258 | |     SingleRule(CbRule),
    | |     ------------------ the second-largest variant contains at least 232 bytes
259 | |     /// If a network rule has more than one specified resource type, one of those types is
...   |
263 | |     SplitDocument(CbRule, CbRule),
    | |     ----------------------------- the largest variant contains at least 464 bytes
264 | | }
    | |_^ the entire enum is at least 464 bytes
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
    = note: `-D clippy::large-enum-variant` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::large_enum_variant)]`
help: consider boxing the large fields to reduce the total size of the enum
    |
263 -     SplitDocument(CbRule, CbRule),
263 +     SplitDocument(CbRule, Box<CbRule>),
```

antonok-edm: Maybe better as `Box<(CbRule, CbRule)>` - but let's avoid either of these for now since it's another breaking API change

atuchin-m: Ok, reverted in https://github.com/brave/adblock-rust/pull/484/commits/2a69988eb6766a696da46aaf2fdbd1f1d011028b

We can fix it later.

antonok-edm (approved):

PR #489, opened 2025-06-20
Use the one flatbuffer to store all lists

atuchin-m:
The PR moves from per-NetworkList flatbuffers to a one (per-Engine).
It doesn't affect the performance metrics, but open possibilities to put cosmetic filters to the same flatbuffer.
It also simplifies the serialization and deserialization code.

Notes:
* It use the original algorithm and structures without futher optimizations. Next time, the diff is big enough. 
* `insert_dup` was dropped, currently it didn't have any effect with flatbuffers (we compared the flatbuffers offsets, not the unique ids). Maybe it makes sense to restore it the future versions.
* `NetworkFilterLists::new` is left as-is for benches and tests to avoid excessive diff.


kdenhartog (commented):
I checked over the unsafe usage and this matches what it was previously. We changed the return type here, but it's not a concern and the implementation matches what it previously was. I'm removing those alerts

src/engine.rs
```diff
@@ -60,15 +63,16 @@ impl Engine {
     /// used with deserialization.
     /// - `optimize` specifies whether or not to attempt to compress the internal representation by
     ///   combining similar rules.
-    pub fn new(optimize: bool) -> Self {
-        let blocker_options = BlockerOptions {
-            enable_optimizations: optimize,
-        };
+    pub fn new(_optimize: bool) -> Self {
+        // TODO: remove _optimize?
```
antonok-edm: I support removing it if it has no effect anymore

src/filters/fb_network.rs
```diff
@@ -212,10 +68,45 @@ impl ExactSizeIterator for FlatPatternsIterator<'_> {
     }
 }
 
+// TODO: do we need another feature for this?
+#[cfg(feature = "unsync-regex-caching")]
+pub(crate) type SharedStateRef = std::rc::Rc<SharedState>;
+#[cfg(not(feature = "unsync-regex-caching"))]
+pub(crate) type SharedStateRef = std::rc::Arc<SharedState>;
```
antonok-edm: Same feature for both is fine, but it would be nice to rename it since it'd no longer be strictly about regex caching

antonok-edm: On that note though, I don't think we actually _need_ any refcounted pointers for this? We could pass the `&SharedState` in as an argument to whatever functions require it, or give `Blocker` a `<'a>` lifetime so it can own a `shared_state: &'a SharedState` field

src/blocker.rs
```diff
@@ -66,26 +67,16 @@ static NO_TAGS: Lazy<HashSet<String>> = Lazy::new(HashSet::new);
...
-
-    pub(crate) enable_optimizations: bool,
-
     // Not serialized
     #[cfg(feature = "unsync-regex-caching")]
     pub(crate) regex_manager: std::cell::RefCell<RegexManager>,
     #[cfg(not(feature = "unsync-regex-caching"))]
     pub(crate) regex_manager: std::sync::Mutex<RegexManager>,
+
+    pub(crate) shared_state: SharedStateRef,
```
boocmp: I think `SharedState` is an unfortunate name, lets discuss it in dm?

src/filters/flat_builder.rs
```diff
@@ -0,0 +1,334 @@
...
+                FilterId::Importants
+            } else if filter.tag.is_some() && !filter.is_redirect() {
+                // `tag` + `redirect` is unsupported for now.
+                FilterId::TaggedFiltersAll
+            } else if (filter.is_redirect() && filter.also_block_redirect())
+                || !filter.is_redirect()
+            {
+                FilterId::Filters
+            } else {
+                continue;
```
boocmp: We don't have this in the original code, do we have hits here? If we got some filters which dont fit any list I'd prefer to know about it

src/filters/flat_builder.rs
```diff
@@ -0,0 +1,334 @@
...
+            };
+
+            builder.add_filter(filter, list_id as u32);
+        }
+
+        builder.finish(if optimize {
```
boocmp: optimize && id != FilterId::RemoveParam as u32 ?

src/filters/flat_builder.rs
atuchin-m: the original code skips unmatched filters the same way:
https://github.com/brave/adblock-rust/blob/master/src/blocker.rs#L453-L457

src/filters/flat_builder.rs
atuchin-m: it was my initial attempt, but we can't use lambas with a capture as fn (like ` fn(u32) -> bool)`)

src/filters/fb_network.rs
atuchin-m: That sounds logical, but the issue is that `Engine` (the primary owner of `shared_state`) also own a Blocker instance (that also stores `shared_state`).

A class member cannot use the same lifetime as another class member without stuff like https://docs.rs/ouroboros/latest/ouroboros/attr.self_referencing.html
Without using `Rc` Blocker will became `Blocker<'a>` with a limited lifetime and can't be stored as a member of Engine, which results in major changes in the codebase.

src/filters/flat_builder.rs
boocmp: Oh, now I see

src/engine.rs
atuchin-m: I believe we should remove it.
The way I see it it did have effect before: we lose ability to add filters one by one sometime ago.
When we called `from_rules_parametrised` - we ignored the value, stored in new()

atuchin-m: `new(..)` is removed in favor of `default()`

src/blocker.rs
atuchin-m: renamed to `FilterDataContext`

src/filters/fb_network.rs
atuchin-m: the feature renamed to `single-thread`

src/filters/flat_builder.rs
```diff
@@ -0,0 +1,337 @@
+//! Builder for creating flatbuffer with serialized engine.
```
boocmp: May be rename it to `fb_builder.rs` ?

boocmp (approved):

src/filters/flat_builder.rs
atuchin-m: we can. @antonok-edm WDYT?

boocmp: I'm asking because we have `fb_network.rs` and `flat_filter_map.rs` and I suppose that fb for `flatbuffers` and `flat` for flat containers

src/filters/flat_builder.rs
```diff
@@ -0,0 +1,337 @@
...
+
+            builder.add_filter(filter, list_id as u32);
+        }
+
+        builder.finish(if optimize {
+            // Don't optimize removeparam, since it can fuse filters without respecting distinct
+            |id: u32| id != FilterId::RemoveParam as u32
+        } else {
+            |_| false
+        })
```
antonok-edm: what's the reason to pass this as a lambda rather than moving the check inside `finish()`?

src/lists.rs
```diff
@@ -231,6 +231,19 @@ impl FilterSet {
         }
     }
 
+    // Used in benchmarks to avoid parsing the rules twice.
+    pub fn new_with_rules(
+        network_filters: Vec<NetworkFilter>,
+        cosmetic_filters: Vec<CosmeticFilter>,
+        debug: bool,
+    ) -> Self {
```
antonok-edm: if it's only for benchmarks, let's `#[doc(hidden)]`

src/filters/flat_builder.rs
antonok-edm: see my other comment, but for future reference: https://doc.rust-lang.org/std/keyword.move.html

src/filters/fb_network.rs
```diff
@@ -212,10 +68,47 @@ impl ExactSizeIterator for FlatPatternsIterator<'_> {
...
+        // Reconstruct the unique_domains_hashes_map from the flatbuffer data
+        let root = memory.root();
+        let mut unique_domains_hashes_map: HashMap<crate::utils::Hash, u32> = HashMap::new();
+        for (index, hash) in root.unique_domains_hashes().iter().enumerate() {
+            unique_domains_hashes_map.insert(hash, index as u32);
+        }
+        FilterDataContextRef::new(Self {
+            memory,
+            unique_domains_hashes_map,
+        })
```
antonok-edm, line 104: not a concern for this particular PR, but does it make sense to hold data that needs to be "reconstructed" in a separate buffer? Then we don't necessarily need to hold both copies in memory.

src/engine.rs
```diff
@@ -110,14 +103,15 @@ impl Engine {
...
-        let blocker_options = BlockerOptions {
-            enable_optimizations: optimize,
-        };
+        let memory = FlatBufferBuilder::make_flatbuffer(network_filters, optimize);
+
+        let filter_data_context = FilterDataContext::new(memory);
 
         Self {
-            blocker: Blocker::new(network_filters, &blocker_options),
+            blocker: Blocker::from_context(filter_data_context.clone()),
```
antonok-edm, line 111: Convention is to explicitly use `Rc::clone(x)` or `Arc::clone(x)` rather than `x.clone()` to make it clear that it's just a refcount increase.

https://doc.rust-lang.org/book/ch15-04-rc.html#using-rct-to-share-data

In these instances it depends on the feature config, so I'd suggest `FilterDataContextRef::clone(x)` instead

src/filters/flat_builder.rs
antonok-edm: yes, sounds good to me

src/filters/fb_network.rs
antonok-edm: My thinking is it could be organized like this minimal example:

```rust
type FilterDataContextRef = [u8; 32];

struct Engine {
    filter_data_ref: FilterDataContextRef,
}

impl Engine {
    fn new() -> Self {
        let filter_data_ref = [0u8; 32];
        Self {
            filter_data_ref,
        }
    }

    fn blocker<'a>(&'a self) -> Blocker<'a> {
        Blocker {
            filter_data_ref: &self.filter_data_ref,
        }
    }

    fn check_network_request(&self) -> bool {
        self.blocker().check()
    }
}

struct Blocker<'a> {
    filter_data_ref: &'a FilterDataContextRef,
}

impl<'a> Blocker<'a> {
    fn check(&self) -> bool {
        true
    }
}
```

Key point is `Engine` owns all the data and has the `blocker()` method to produce a scoped convenience struct that can be used for the current network blocking methods without moving too much code around.

The only other fields on `Blocker` are:
- `regex_manager`, which also makes sense to keep on the top-level `Engine` so that it may be used for cosmetic filtering as needed (some of uBO's newer syntax features could benefit from it)
- `tags_enabled`, which I don't have strong preferences about since I'd like to get rid of it anyways

src/filters/fb_network.rs
atuchin-m: The problem is that stored `unique_domains_hashes` is index => ShortHash (basically), but we need ShortHash => index.
Maybe it makes sense to store additional mapping (despite it will eat some small amount memory).
Well, I believe we need to dedup and index not only domains, but the other strings, so let's postpone this for some time.

atuchin-m:
>one more general thing - could we do these in a 0.11.x branch rather than in master? That way it's a little bit easier to maintain the history of breaking changes vs patch releases
We can in theory. But the workflow isn't clear to me.
Create 0.11.x, make a PR here, merge it, then merge 0.11.x to master?
The approach to merge everything in `master` first is more clear to me, also in terms of tracking perf (a linear history).

src/lists.rs
atuchin-m: Done

src/filters/flat_builder.rs
atuchin-m: Heh, the initial idea was to make 2 layers: for the storage and for the logic how and what.
But in fact that parts of code is strongly connected to each other.
Changed to just `bool`.

src/engine.rs
atuchin-m: Done, thanks.

PR #498, opened 2025-08-11
Flat containers

boocmp:
This is a preparation PR for moving cosmetic filters into flatbuffers.
1. Replaced `FlatFilterMap` with `FlatMultiMapView` - a common container that will also be used in `CosmeticFilterCache`
2. Introduced `FlatSetView` (currently unused), which will be used in `CosmeticFilterCache`
3. Moved flatbuffers utilities to a `flatbuffers` folder


src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,88 @@
...
+use flatbuffers::{Follow, Vector};
+
+/// A map-like container that uses flatbuffer references.
+/// Provides O(log n) lookup time using binary search on the sorted index.
+pub(crate) struct FlatMultiMapView<'a, I: Ord + Copy, V, Idx>
+where
+    Idx: Indexable<I>,
+    V: Follow<'a>,
+{
+    index: Idx,
```
atuchin-m: let's use reserve `index` name only for usize variables. Let's use `keys` here.

src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,88 @@
+use std::marker::PhantomData;
+
+use crate::flatbuffers::containers::indexable::Indexable;
+use flatbuffers::{Follow, Vector};
+
+/// A map-like container that uses flatbuffer references.
+/// Provides O(log n) lookup time using binary search on the sorted index.
+pub(crate) struct FlatMultiMapView<'a, I: Ord + Copy, V, Idx>
```
atuchin-m: Let's add a comment what with the meaning of the params.

src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,88 @@
...
+{
+    type Item = (usize, <V as Follow<'a>>::Inner);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_index < self.indexes.len() {
+            if self.indexes.get(self.current_index) != self.key {
+                return None;
+            }
+            let index = self.current_index;
+            let filter = self.values.get(self.current_index);
```
atuchin-m:
```suggestion
            let filter = self.values.get(index);
```

src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn get(&self, key: I) -> FlatMultiMapViewIterator<'a, I, V, Idx>
+    where
+        Idx: Clone,
+    {
+        let start = self.index.partition_point(|x| *x < key);
+        FlatMultiMapViewIterator {
+            current_index: start,
+            key,
+            indexes: self.index.clone(),
```
atuchin-m: Let's add a comment why we use `.clone()` here.

src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,88 @@
...
+        Self {
+            index,
+            values,
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn get(&self, key: I) -> FlatMultiMapViewIterator<'a, I, V, Idx>
+    where
+        Idx: Clone,
```
atuchin-m: should we move this to the top?

src/flatbuffers/containers/indexable.rs
```diff
@@ -0,0 +1,61 @@
+use flatbuffers::{Follow, Vector};
+
+pub(crate) trait Indexable<I> {
```
atuchin-m: Add some comments.

Cargo.toml
```diff
@@ -104,6 +104,7 @@ resource-assembler = []
 
 [lints.clippy]
 len_zero = "allow"
+uninlined_format_args = "allow"
```
atuchin-m: should we put this to a dedicated PR?
with live redirect tests disabling

src/flatbuffers/containers/sorted_index.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn partition_point<F>(&self, mut predicate: F) -> usize
+    where
+        F: FnMut(&T::Inner) -> bool,
+    {
+        debug_assert!(self.iter().is_sorted());
+
+        let mut left = 0;
+        let mut right = self.len();
+
+        while left < right {
```
atuchin-m, line 61: not clear why we need two manually written `partition_point`

atuchin-m: upd: looks like PartitionPoint is leftover, should be deleted

Cargo.toml
antonok-edm: was handled by #502

tests/live.rs
```diff
@@ -217,6 +217,7 @@ fn check_live_from_filterlists() {
 
 #[cfg(feature = "resource-assembler")]
 #[test]
+#[ignore = "issues/499"]
```
antonok-edm, line 233: also fixed by #503, this test should not be ignored

src/flatbuffers/containers/flat_multimap.rs
```diff
@@ -0,0 +1,86 @@
+use std::marker::PhantomData;
+
+use crate::flatbuffers::containers::sorted_index::SortedIndex;
+use flatbuffers::{Follow, Vector};
+
+/// A map-like container that uses flatbuffer references.
+/// Provides O(log n) lookup time using binary search on the sorted index.
+/// I is a key type, Keys is specific container of keys, &[I] for fast indexing (u32, u64)
+/// and flatbuffers::Vector<I> if there is no conversion from Vector (str) to slice.
```
antonok-edm, line 9: Since we're accumulating several of these flatbuffer utility structs, I'm wondering if it makes sense to create and publish a new separate crate just for those? I imagine others in the Rust community could find them useful too.

No need to do so here, but something to consider.

antonok-edm (approved):
approved assuming the above are addressed

src/flatbuffers/containers/flat_multimap.rs
atuchin-m: That could be an option once we finally stabilize the interface and merge it to master.

tests/live.rs
atuchin-m: the bad thing that we have 2 branches now.
We fixed this into `master`, but this PR related only to `0.11.x`.
I've manually cherry-picked and pushed https://github.com/brave/adblock-rust/pull/502 and https://github.com/brave/adblock-rust/pull/503 to `0.11.x` to address this.

I believe we need to clarify our branching strategy here.

atuchin-m (approved):

PR #507, opened 2025-08-14
Flat container builders

atuchin-m:


src/flatbuffers/unsafe_tools.rs
```diff
@@ -3,8 +3,7 @@
 use crate::filters::fb_network::flat::fb;
 
 // Minimum alignment for the beginning of the flatbuffer data.
-// Should be 4 while we support armv7 and x86_32.
-const MIN_ALIGNMENT: usize = 4;
```
antonok-edm, line 7: no longer supporting 32-bit architectures at all?

antonok-edm (approved):

src/flatbuffers/unsafe_tools.rs
atuchin-m: we do. in case the point isn't aligned to 4 bytes it will be aligned manually.

PR #508, opened 2025-08-14
Fix mismatched_lifetime_syntaxes

atuchin-m:
The PR addresses a new Rust warning `mismatched_lifetime_syntaxes`


src/filters/fb_network.rs
```diff
@@ -15,7 +15,14 @@ use crate::regex_manager::RegexManager;
 use crate::request::Request;
 use crate::utils::{Hash, ShortHash};
 
-#[allow(dead_code, clippy::all, unused_imports, unsafe_code)]
+#[allow(unknown_lints)]
```
atuchin-m, line 18: This is necessary for old Rust versions to support disabling `mismatched_lifetime_syntaxes`

PR #514, opened 2025-08-25
Cosmetic filter fb

atuchin-m:
The PR introduces new structures to store cosmetic filters in flatbuffer.
* the algorithms to sort and apply rules shouldn't be touched, only storage level is changed.
* CosmeticFilterCache is now a view for a flatbuffer data.
* Old storage layer (via serde) is removed, the version is now stored in the flatbuffer,
* Another container 'FlatMap' is introduced + tests
* Most host-specific rules are stored in a single FlatMap (domain_hash => `HostnameSpecificRules`).  Although, the most common rule kinds are stored as a dedicated multi-maps to save memory
* a code to build flatbuffer structure is moved to dedicated files.

Perf impact:
* improve memory usage by 43%
* improves loading .dat file 3x times
* a non-important `rule-match-first-request/brave-list` regression, the absolute number ~1ms is still totally fine for the first request.
* a potentially bad regression with `cosmetic-class-id-match/brave-list`.A new code uses O(log N) binany search instead of O(1) HashSomething lookup. For large number of selectors (~1000) that can make sense. We're going to improve this in a dedicated PR.


src/cosmetic_filter_cache.rs
```diff
@@ -191,34 +166,40 @@ impl CosmeticFilterCache {
     ) -> Vec<String> {
         let mut selectors = vec![];
 
+        let cs = self.filter_data_context.memory.root().cosmetic_filters();
```
boocmp: `cf` ?, or even `cosmetic_filters`

src/blocker.rs
```diff
@@ -440,11 +440,13 @@ impl Blocker {
...
     ) -> Self {
-        use crate::filters::{fb_builder::make_flatbuffer, fb_network::FilterDataContext};
+        use crate::engine::Engine;
+        use crate::FilterSet;
 
-        let memory = make_flatbuffer(network_filters, options.enable_optimizations);
-        let filter_data_context = FilterDataContext::new(memory);
-        Self::from_context(filter_data_context)
+        let mut filter_set = FilterSet::new(true);
+        filter_set.network_filters = network_filters;
```
antonok-edm, line 447: It's a bit odd to have `Blocker::new()` build a FilterSet, then manually move network filters into the set, then build an `Engine`, and finally make the blocker from the engine. Understood this is test only, but perhaps we should consider a different constructor method for the tests.

src/data_format/mod.rs
```diff
@@ -1,111 +0,0 @@
...
-pub(crate) mod utils;
-
-use crate::cosmetic_filter_cache::CosmeticFilterCache;
-use crate::flatbuffers::unsafe_tools::VerifiedFlatbufferMemory;
-use crate::network_filter_list::NetworkFilterListParsingError;
-
-/// Newer formats start with this magic byte sequence.
-/// Calculated as the leading 4 bytes of `echo -n 'brave/adblock-rust' | sha512sum`.
-const ADBLOCK_RUST_DAT_MAGIC: [u8; 4] = [0xd1, 0xd9, 0x3a, 0xaf];
-const ADBLOCK_RUST_DAT_VERSION: u8 = 1;
```
antonok-edm: could we keep the byte sequence and version at the start of the serialized file? Brave iOS is currently using it.

src/engine.rs
```diff
@@ -51,16 +57,18 @@ pub struct Engine {
     filter_data_context: FilterDataContextRef,
 }
 
+const ADBLOCK_FLATBUFFER_VERSION: u32 = 1;
```
antonok-edm: should be incremented

atuchin-m: Because it a new format, we can start with 1 here.

src/data_format/mod.rs
atuchin-m: Do we use that magic somewhere outside adblock-rust to identify the file?

Now the while file content is a flatbuffer and we verify the format before file using, so it looks like we don't need a extra magic.

src/blocker.rs
atuchin-m: Yeah, that test ideally these test contractors should be removed in favor of building the engine directly (or another high level test method).
Although, preserving them allow to reduce the diff in the PR (we don't need to rewrite a lot of tests). So I suggest to change it in another PR.

P.S. We can avoid using Engine completely because it's a flatbuffer root table. And we need to get serialized flatbuffer data.

src/flatbuffers/fb_network_filter.fbs
```diff
@@ -29,17 +29,91 @@ table NetworkFilterList {
   filter_map_values: [NetworkFilter] (required);
 }
 
+// A table to store the most host-specific cosmetic rules.
+// Although, the most common kind of rule (see hostname_inject_script_*
+// and hostname_hide_*) are stored separately to save memory.
```
antonok-edm: Just noting the `///` doc comments are applied inconsistently

src/cosmetic_filter_cache.rs
```diff
@@ -7,16 +7,23 @@
...
+//! To build `CosmeticFilterCache`, use `CosmeticFilterCacheBuilder`.
 
-use crate::filters::cosmetic::{
-    CosmeticFilter, CosmeticFilterAction, CosmeticFilterMask, CosmeticFilterOperator,
-};
+use crate::cosmetic_filter_utils::decode_script_with_permission;
+#[cfg(test)]
+use crate::filters::cosmetic::CosmeticFilter;
+use crate::filters::cosmetic::{CosmeticFilterAction, CosmeticFilterOperator};
+use crate::filters::fb_network::FilterDataContextRef;
```
antonok-edm: probably worth renaming or moving out of `fb_network` now that it's used by cosmetic filters

src/cosmetic_filter_cache_builder.rs
```diff
@@ -0,0 +1,247 @@
+//! Provides API to prepare and serialize cosmetic filter rules to a flatbuffer.
+//! To build the struct, use `CosmeticFilterCacheBuilder`.
+//! To use the serialized rules, use `CosmeticFilterCache`.
+
+use crate::cosmetic_filter_cache::ProceduralOrActionFilter;
+use crate::cosmetic_filter_utils::SpecificFilterType;
+use crate::cosmetic_filter_utils::{encode_script_with_permission, key_from_selector};
+use crate::filters::cosmetic::{CosmeticFilter, CosmeticFilterMask, CosmeticFilterOperator};
+use crate::filters::fb_network::flat::fb;
```
antonok-edm: another `fb_network` used by cosmetic filters

src/data_format/mod.rs
antonok-edm: When the iOS client is updated to use adblock-rust v0.11.x, it will read the older file and try to interpret it as a flatbuffer rather than looking at the first bytes and knowing to exit immediately because of a version mismatch.

If we're lucky that file will have a deserialization error, but in the worst case it may parse and be full of junk data. Keeping the separate format header will guarantee future clients are always able to early-exit even if we ever switch to something other than flatbuffers later on.

atuchin-m: Verifying a flat buffer is quick and the migration happens once a few months. I don't think it makes sense to save some CPU cycles here.

As for correctness: theoretically it's possible to have a broken file that will be interpreted as valid flatbuffer. The same thing for the old .dat: a partially written file will pass magic header verification.

I would add some kind of CRC sum instead (to be sure that the file is consistent).
If you want to preserve magic number: let's keep it, but as the part of a flatbuffer.

src/cosmetic_filter_cache.rs
atuchin-m: yep, it makes sense to move `FilterDataContext*`

atuchin-m: done

src/flatbuffers/fb_network_filter.fbs
atuchin-m: fixed, thanks

src/data_format/mod.rs
atuchin-m: added seahash as a checksum instead.

src/cosmetic_filter_cache_builder.rs
atuchin-m: Fixed

src/data_format/mod.rs
antonok-edm: checksum doesn't address my concern, let me try to explain the consequences a bit more clearly:

There are currently millions of active iOS installations of Brave Browser with a cached file of the format `[ADBLOCK_RUST_DAT_MAGIC[0..4], ADBLOCK_RUST_DAT_VERSION, <SomeMessagePackPayload>[0..]]`. This data wasn't serialized as a flatbuffer, so it is _unlikely_ to deserialize into a valid flatbuffer according to our new schema here. However, "unlikely" isn't good enough for a file that is read into memory at startup and could cause irrecoverable browser crash loops<sup>1</sup> if internal invariants about the data structures are not upheld properly.

Even worse, we cannot empirically verify parsing incompatibility on a historical snapshot of that cached file because each browser install may have different lists enabled and may have even serialized this file using any previous version of adblock-rust.

If you'd like to go through the effort to read the specs for both flatbuffer and messagepack's internal binary representations, as well as their respective (current and historical) library code to rule out possible implementation quirks, then you _may_ be able to prove that it is impossible for a previous version of adblock-rust to have serialized any `Engine` into a payload that this new flatbuffer implementation will be able to deserialize without errors. That would make me ~70% comfortable with leaving out such a header check.

Alternatively, we could keep the header and be 100% sure it will be respected. This particular use case is exactly why it was added originally.

<br>

<sup>1</sup>: _there is [ongoing](https://docs.google.com/document/d/1PE2wk1gBuxsEFljmgDwIot2NIJDq9BneDnOyXXissXY/edit?usp=sharing) (but currently incomplete) work to isolate this in a separate process to at least make it recoverable at runtime._

antonok-edm: FWIW the checksum is probably a good idea too in the event of disk corruption, but it's not a complete solution to that migration problem.

atuchin-m: I got your concerns and returned the magic header.

I would say the situation you're describing is almost impossible, because the current data file have to pass 3 independent checks:
1. the flatbuffer format verification(many bytes, >8)
2. version (4 bytes)
3. checksum (8 bytes) 

But there is nothing wrong with one extra check for sure.

P.S. Chrome uses a just a checksum in about the same situation: https://chromium-review.googlesource.com/c/chromium/src/+/1093152

src/data_format/mod.rs
```diff
@@ -4,92 +4,55 @@
...
-pub(crate) mod utils;
-
-use crate::cosmetic_filter_cache::CosmeticFilterCache;
-use crate::flatbuffers::unsafe_tools::VerifiedFlatbufferMemory;
-use crate::network_filter_list::NetworkFilterListParsingError;
+//!
+//! The current .dat file format:
+//! 1. magic (4 bytes)
+//! 2. seahash of the data (8 bytes)
+//! 3. data (the rest of the file)
```
antonok-edm: we still need the version in the header, in the case of the hash starting with a `1` byte

atuchin-m: Why? For flatbuffers we have the version field in .fbs, that's easy to store/check.

antonok-edm: for the same reason in https://github.com/brave/adblock-rust/pull/514#discussion_r2345543495 - the old code expects that a magic header followed by a `1` byte is a valid payload and so we should guard against it

atuchin-m: returned `version` to .dat header.

src/data_format/mod.rs
```diff
@@ -4,92 +4,67 @@
...
 const ADBLOCK_RUST_DAT_MAGIC: [u8; 4] = [0xd1, 0xd9, 0x3a, 0xaf];
-const ADBLOCK_RUST_DAT_VERSION: u8 = 1;
 
-#[derive(Debug)]
-pub enum SerializationError {
-    RmpSerdeError(rmp_serde::encode::Error),
-}
+/// The version of the data format.
+/// If the data format version is incremented, the data is considered as incompatible.
+const ADBLOCK_FLATBUFFER_VERSION: u8 = 1;
```
antonok-edm: and again this should now be incremented to 2, otherwise good

atuchin-m: done

PR #515, opened 2025-08-29
Update `selectors` crate to v0.26

antonok-edm:
The purpose of this update is to align the `selectors` version with what is currently used for other dependencies in Brave Browser, however this actually fixes a subtle issue relating to escaping special characters that occurs in 3 filter rules from our default lists. Tests added and updated to reflect the fix.

### DIFF
--- a/tests/unit/engine.rs
+++ b/tests/unit/engine.rs
@@ -219,7 +219,7 @@ mod tests {
         let data = engine.serialize().unwrap();
 
         let expected_hash = if cfg!(feature = "css-validation") {
-            7254547691107602751
+            6556399856993017327
         } else {
             4130628479730907288
         };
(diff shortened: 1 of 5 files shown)


tests/unit/engine.rs
```diff
@@ -219,7 +219,7 @@ mod tests {
         let data = engine.serialize().unwrap();
 
         let expected_hash = if cfg!(feature = "css-validation") {
-            7254547691107602751
+            6556399856993017327
```
antonok-edm, line 222: There is no format change here, this checksum update just reflects the 3 fixed filters from `brave-main-list.txt`.

boocmp (approved):

cdesouza-chromium (approved):
LGTM

PR #517, opened 2025-09-11
Optimize resource storage memory

atuchin-m:
The PR supports decoding resources before storing it. This optimize memory usage of `ResourceStorage` by ~30%.
* add a new `ResourceImpl` internal struct to store decoded resources
* validate each resource during initial loading
* support encoding a resource back to base64 when we need this (`get_redirect_resource()`)

### DIFF
--- a/src/resources/resource_storage.rs
+++ b/src/resources/resource_storage.rs
@@ -34,7 +34,7 @@ impl ResourceContent {
 #[derive(Clone)]
 /// A internal representation of a Resource to store. Stores the content
 /// in the decoded form to use less memory.
-/// See Resource for details.
+/// See [Resource] for details
 struct ResourceImpl {
     name: String,
     kind: ResourceType,


src/resources/resource_storage.rs
```diff
@@ -9,11 +9,45 @@ use thiserror::Error;
...
+    fn raw_from_base64(base64: &str) -> Result<Self, AddResourceError> {
+        let decoded = BASE64_STANDARD.decode(base64)?;
+        Ok(Self::Raw(decoded))
+    }
+}
+
+#[derive(Clone)]
+/// A internal representation of a Resource to store. Stores the content
+/// in the decoded form to use less memory.
+/// See Resource for details.
```
antonok-edm: Let's take advantage of rustdoc's inter-comment linking with square brackets, e.g.:

`/// See [Resource] for details`

atuchin-m: thanks, done.

PR #519, opened 2025-09-19
Allow to share resources, expose more debug info

atuchin-m:
The PR changes the public API to:
* provide ability to share resources between the engines (that will save about 2 MB of memory in the browser process after it is supported in brave-core)
* expose more debug info to see the more detailed memory usage.

### DIFF
--- a/tests/unit/engine.rs
+++ b/tests/unit/engine.rs
@@ -238,6 +238,38 @@ mod tests {
     }
 
     #[test]
+    fn resource_storage_sharing() {
+        let resources = ResourceStorage::from_resources([Resource::simple(
+            "refresh-defuser.js",
+            MimeType::ApplicationJavascript,
+            "refresh-defuser",
+        )]);
+        let mut engine1 =
+            Engine::from_rules(["domain1.com##+js(refresh-defuser)"], Default::default());
+        let mut engine2 =
+            Engine::from_rules(["domain2.com##+js(refresh-defuser)"], Default::default());
+        engine1.use_resource_storage(ResourceStorageRef::clone(&resources));
+        engine2.use_resource_storage(ResourceStorageRef::clone(&resources));
+
+        fn wrap_try(scriptlet_content: &str) -> String {
+            format!("try {{\n{}\n}} catch ( e ) {{ }}\n", scriptlet_content)
+        }
+
+        assert_eq!(
+            engine1
+                .url_cosmetic_resources("https://domain1.com")
+                .injected_script,
+            wrap_try("refresh-defuser")
+        );
+        assert_eq!(
+            engine2
+                .url_cosmetic_resources("https://domain2.com")
+                .injected_script,
+            wrap_try("refresh-defuser")
+        );
+    }
+
+    #[test]
     fn redirect_resource_insertion_works() {
         let mut engine = Engine::from_rules(
             ["ad-banner$redirect=nooptext", "script.js$redirect=noop.js"],


src/engine.rs
```diff
@@ -246,9 +258,13 @@ impl Engine {
...
-    #[cfg(feature = "regex-debug-info")]
-    pub fn get_regex_debug_info(&self) -> crate::regex_manager::RegexDebugInfo {
-        self.blocker.get_regex_debug_info()
+    pub fn get_debug_info(&self) -> EngineDebugInfo {
+        EngineDebugInfo {
+            #[cfg(feature = "regex-debug-info")]
+            regex_debug_info: self.blocker.get_regex_debug_info(),
+            flatbuffer_size: self.filter_data_context.memory.data().len(),
+            resources_total_length: self.resources.resources_total_length(),
+        }
```
antonok-edm, line 267: can we keep the method and struct definition behind the (renamed) `debug-info` flag?

src/resources/resource_storage.rs
```diff
@@ -9,6 +9,14 @@ use thiserror::Error;
 
 use super::{PermissionMask, Resource, ResourceType};
 
+/// A ref-counted reference to a [`ResourceStorage`].
+#[cfg(feature = "single-thread")]
+pub type ResourceStorageRef = std::rc::Rc<ResourceStorage>;
+
+/// A ref-counted reference to a [`ResourceStorage`].
+#[cfg(not(feature = "single-thread"))]
+pub type ResourceStorageRef = std::sync::Arc<ResourceStorage>;
```
antonok-edm, line 18: Let's implement this using traits instead; it will give us more flexibility to e.g. use an on-disk cache in brave-core and have better devtools diagnostics in the future.

More or less like this:
```rust
/// Customizable backend for [Resource] storage.
/// Custom implementations could be used to enable (for example) sharing of resources between multiple [Engine]s, an on-disk backend, or special caching behavior.
pub trait ResourceStorageBackend {
    /// Retrieves the resource with the associated name or alias
    fn get_resource(&self, resource_ident: &str) -> Option<Resource>;
}

/// Default implementation of `ResourceStorageBackend` that stores all resources in memory.
pub struct InMemoryResourceStorageBackend {
    /// Stores each resource by its canonical name
    resources: HashMap<String, Resource>,
    /// Stores mappings from aliases to their canonical resource names
    aliases: HashMap<String, String>,
}

impl ResourceStorageBackend for InMemoryResourceStorageBackend {
    fn get_resource(&self, resource_ident: &str) -> Option<Resource> {
        // ...the current `get_internal_resource` method implementation (note `&Resource` -> `Resource`)
    }
}

impl InMemoryResourceStorageBackend {
    pub fn from_resources(resources: impl IntoIterator<Item = Resource>) -> Self {
        // ...the current `ResourceStorage::from_resources` method
    }
}

// ...

/// Unified resource storage for both redirects and scriptlets.
pub struct ResourceStorage {
   backend: Box<ResourceStorageBackend>,
}

impl ResourceStorage {
    pub fn use_backend<T>(&mut self, backend: T) where T: ResourceStorageBackend + 'static {
        self.backend = Box::new(backend);
    }
    // ...replace all `self.get_internal_resource` calls with `backend.get_resource`
}
```

Dropping support for individual `add_resource` calls is ok with me. `Engine` can keep the existing `use_resources` method to create an `InMemoryResourceStorageBackend` with the provided resources, as well as a new method `use_resources_backend`.

src/engine.rs
atuchin-m: ok, let's rename the feature to `debug-info` and guard by it.

src/resources/resource_storage.rs
atuchin-m: The particular task I'm trying to solve it **sharing** the impl between the engines.

It's sounds like you suggest changing the interfaces in another place.
Right now we have only one `ResourceStorage` impl and no plans to support another. I have no idea what to put into `InMemoryResourceStorageBackend`. Can you elaborate on this?

antonok-edm: `InMemoryResourceStorageBackend` is identical to the current implementation. What I propose with the `ResourceStorageBackend` trait is that we can use it to define a new brave-core specific implementation that supports sharing between engines and (eventually) on-disk caching. We can put the implementation of that in brave-core's `adblock` binding layer since we will need tighter integration with Chromium's FS operations; it doesn't need to live in `adblock-rust` itself.

```rust
/// To be wrapped in [Rc] for shared access across engines
struct BraveCoreResourceStorageInner {
    /// Stores each resource by its canonical name
    resources: HashMap<String, Resource>,
    /// Stores mappings from aliases to their canonical resource names
    aliases: HashMap<String, String>,
}

#[derive(Clone)]
struct BraveCoreResourceStorage {
    shared_storage: Rc<BraveCoreResourceStorageInner>,
}

impl ResourceStorageBackend for BraveCoreResourceStorage {
    fn get_resource(&self, resource_ident: &str) -> Option<Resource> {
        // ...use `self.shared_storage.resources` and `self.shared_storage.aliases`
    }
}
```

src/engine.rs
atuchin-m: done

src/resources/resource_storage.rs
atuchin-m: The current conception assumes that the browser should just make `ResourceStorage` and pass it to the both engines.

> trait is that we can use it to define a new brave-core specific implementation that supports sharing between engines and (eventually) on-disk caching.

I don't have such plans right now. Also, on-disk representation takes more memory because of base64: an extra work is needed to process it or change the backend code to storage. If you have a conception how to deal with it, it would be nice to see a PR.
I would say we first need to focus on 20MB+ flatbuffer caching rather that relatively small resources.

In conclusion, I don't want to make multiple `get_resource`.

The only thing I'm generalizing here is getting the current `ResourceStorage` impl.
If you really want it, we can do it via traits: `ResourceStorageGetter` with two impls `SimpleResourceStorage` and `SharableResourceStorageGetter`.

atuchin-m: okay, I made a dedicated PR with the part with extra debug info: https://github.com/brave/adblock-rust/pull/527
At least we can merge this part and continue discussion about this.
I've also added the test to show how the resources can be shared from the b-c perspective.

PR #521, opened 2025-09-25
ci: Fix rust toolchain

mschfh:


rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "stable"
```
atuchin-m, line 2: Let's also change this to "1.90"

atuchin-m (approved):

rust-toolchain.toml
atuchin-m: Will do it in another PR

PR #526, opened 2025-10-01
Allow custom resource storage backends, expose more debug info 

antonok-edm:
supersedes #519 

breaking API changes from this PR:

- Added `ResourceStorageBackend`. This trait can be implemented to use customized storage backend logic, including sharing between multiple engines, caching from disk, ability to dynamically add/remove resources at runtime, collecting metrics, etc.
- Added `InMemoryResourceStorage` as a port of the original `ResourceStorage` backend.
- Removed `Engine::add_resource`. If you still need the ability to add individual resources at runtime, use a custom `ResourceStorageBackend` instead.
- The `regex-debug-info` feature has been named to `debug-info`, as it now also enables a new API to get the size of the `Engine`'s internal flatbuffer at runtime.

### DIFF
--- a/tests/unit/resources/resource_storage.rs
+++ b/tests/unit/resources/resource_storage.rs
@@ -805,3 +805,71 @@ mod scriptlet_storage_tests {
         assert_eq!(resources.get_scriptlet_resources([("test, 1", PERM01), ("test-wrapper, 2", PERM01), ("shared, 3", Default::default())]), "permissioned\na\ncommon\nb\nfunction test() {}\nfunction testWrapper() { test(arguments) }\nfunction shared() { }\ntry {\ntest(\"1\")\n} catch ( e ) { }\ntry {\ntestWrapper(\"2\")\n} catch ( e ) { }\ntry {\nshared(\"3\")\n} catch ( e ) { }\n");
     }
 }
+
+#[cfg(all(test, feature = "single-thread"))]
+mod shared_storage_tests {
+    use super::super::*;
+    use crate::resources::MimeType;
+
+    use std::rc::Rc;
+
+    /// To be wrapped in [Rc] for shared access across engines.
+    struct BraveCoreResourceStorageInner {
+        /// Stores each resource by its canonical name
+        resources: HashMap<String, Resource>,
+    }
+
+    #[derive(Clone)]
+    struct BraveCoreResourceStorage {
+        shared_storage: Rc<BraveCoreResourceStorageInner>,
+    }
+
+    impl ResourceStorageBackend for BraveCoreResourceStorage {
+        fn get_resource(&self, resource_ident: &str) -> Option<Resource> {
+            self.shared_storage.resources.get(resource_ident).cloned()
+        }
+    }
+
+    #[test]
+    fn share_resources() {
+        let shared_storage = Rc::new(BraveCoreResourceStorageInner {
+            resources: HashMap::from_iter([(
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
+      # Can be removed once https://github.com/actions/runner-images/pull/13076 is released
```
atuchin-m, line 32: Let's remove this to make the diff clear.
I've pushed the recent fixes from `master` to `0.11.x`

src/flatbuffers/unsafe_tools.rs
```diff
@@ -17,12 +17,12 @@ pub fn fb_vector_to_slice<T>(vector: flatbuffers::Vector<'_, T>) -> &[T] {
         // We can't use T with the size more than MIN_ALIGNMENT.
         // Since the beginning of flatbuffer data is aligned to that size,
         // the alignment of the data must be a divisor of MIN_ALIGNMENT.
-        assert!(MIN_ALIGNMENT % std::mem::size_of::<T>() == 0);
+        assert!(MIN_ALIGNMENT.is_multiple_of(std::mem::size_of::<T>()));
```
atuchin-m, line 20: the same, let's revert unrelated changes.

src/resources/resource_storage.rs
```diff
@@ -10,14 +10,124 @@ use thiserror::Error;
 use super::{PermissionMask, Resource, ResourceType};
 
 /// Unified resource storage for both redirects and scriptlets.
-#[derive(Default)]
+///
+/// By default, this uses an in-memory storage implementation, however this can be changed using
+/// a custom [ResourceStorageBackend] if desired.
 pub struct ResourceStorage {
+    #[cfg(not(feature = "single-thread"))]
+    backend: Box<dyn ResourceStorageBackend + Sync + Send>,
```
atuchin-m, line 18: hm, I'm not strongly again `Box<dyn ..>` it makes the code less clear and may have some negative perf effect (if it's a hot path)

src/resources/resource_storage.rs
```diff
@@ -10,14 +10,124 @@ use thiserror::Error;
...
+            self.resources.get(canonical_name)
+        } else {
+            None
+        };
+
+        resource.cloned()
+    }
+}
+
+impl InMemoryResourceStorage {
```
atuchin-m, line 85: That code have major conflicts with `master` with https://github.com/brave/adblock-rust/pull/517.

src/resources/resource_storage.rs
```diff
@@ -10,14 +10,124 @@ use thiserror::Error;
...
+    #[cfg(test)]
+    pub fn in_memory_from_resources(resources: impl IntoIterator<Item = Resource>) -> Self {
+        Self::from_backend(InMemoryResourceStorage::from_resources(resources))
+    }
+}
+
+/// Customizable backend for [Resource] storage.
+/// Custom implementations could be used to enable (for example) sharing of resources between
+/// multiple [crate::Engine]s, an on-disk backend, or special caching behavior.
+pub trait ResourceStorageBackend {
```
atuchin-m, line 57: That trait looks internal to me. Can we don't make it public? I would say the best option is just to provide a few public methods to build `ResourceStorage` without exposing the other struct/traits

PR #528, opened 2025-10-14
Speed up cosmetic filtering

atuchin-m:
The PR fixes cosmetic filtering performance that [regressed](https://github.com/brave/adblock-rust/pull/514#pullrequestreview-3151528529) after refactoring cosmetic filters to flatbuffers (binary search is slower than hash set/map)
That's a dealbreaker for rolling 0.11.x.

The PR:
* Introduce new structures `FlatHash{Set,Map}` that uses open hashing approach. That provide a good balance between performance and memory usage.
* Hasher is used `rustc-hash` (the same version used in Chromium)
* Add new traits to reuse the current code.
* Memory usage is increased (for 0.11.x: 10 MB => 10.5MB, looks acceptable)

~~TODO: add unit tests (if the rest looks fine).~~  Done

### DIFF
--- a/tests/unit/engine.rs
+++ b/tests/unit/engine.rs
@@ -221,7 +221,7 @@ mod tests {
         #[cfg(feature = "debug-info")]
         {
             let debug_info = engine.get_debug_info();
-            let expected_size = 8_527_344_f32;
+            let expected_size = 8_963_552_f32;
             assert!(debug_info.flatbuffer_size >= (expected_size * 0.99) as usize);
             assert!(debug_info.flatbuffer_size <= (expected_size * 1.01) as usize);
         }


src/flatbuffers/containers/flat_serialize.rs
```diff
@@ -27,8 +27,13 @@ pub trait FlatSerialize<'b, B: FlatBuilder<'b>>: Sized {
 
 impl<'b> FlatBuilder<'b> for flatbuffers::FlatBufferBuilder<'b> {
     fn create_string(&mut self, s: &str) -> WIPOffset<&'b str> {
-        self.create_string(s)
+        if s.is_empty() {
+            self.create_shared_string(s)
+        } else {
+            self.create_string(s)
```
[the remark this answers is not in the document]
atuchin-m: Fixed.

src/flatbuffers/containers/fb_index.rs
```diff
@@ -0,0 +1,45 @@
+use flatbuffers::{Follow, Vector};
+
+/// A trait to access indexed data in a flatbuffer.
+/// It has two implementations:
+/// 1. a faster &[I] for slices;
+/// 2. a slower for flatbuffers::Vector<I>, that uses Follow() internally.
+///
+/// Note: it intentally returns values using a copy, because it's faster
```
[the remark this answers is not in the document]
atuchin-m: Fixed.

Cargo.lock
```diff
@@ -1586,6 +1587,12 @@ version = "0.1.24"
...
 
+[[package]]
+name = "rustc-hash"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
+
 [[package]]
 name = "rustc-hash"
 version = "2.1.1"
```
antonok-edm, line 1598: I suppose given that `2.1.1` is already in the dependency tree here, we're probably better off using that single version instead of two different ones?

src/cosmetic_filter_cache_builder.rs
```diff
@@ -56,13 +58,18 @@ impl<'a, B: FlatBuilder<'a>> FlatSerialize<'a, B> for HostnameRule {
     }
 }
 
+#[derive(Default, Clone)]
+struct StringVector {
+    data: Vec<String>,
+}
```
antonok-edm: let's express this as a tuple struct, i.e. `struct StringVector(Vec<String>)`

src/flatbuffers/containers/fb_index.rs
```diff
@@ -0,0 +1,45 @@
...
+/// A trait to access indexed data in a flatbuffer.
+/// It has two implementations:
+/// 1. a faster &[I] for slices;
+/// 2. a slower for flatbuffers::Vector<I>, that uses Follow() internally.
+///
+/// Note: it intentionally returns values using a copy, because it's faster
+/// than by reference.
+pub(crate) trait FbIndex<I> {
+    fn len(&self) -> usize;
+    fn get(&self, index: usize) -> I;
```
antonok-edm, line 16: None of the implementations build in any length checks, so overflowed accesses may panic. Let's make sure that is documented in the trait.

Cargo.lock
atuchin-m: I used `1.1.0` because switching to `2.1.1` increases the benchmark result in 15% 🤷‍♂️ .
I suspect Chromium still uses 1.1.0 for the same reason (so it's in the dependency tree for the browser)

src/cosmetic_filter_cache_builder.rs
atuchin-m: done.

src/flatbuffers/containers/fb_index.rs
atuchin-m: Thanks. I've added comments.
Just to clarify: `get()` does range check every time (that's not always good for perf), but panics if it fails.

Cargo.lock
antonok-edm: hmm, ok. That sounds like a good thing to document in the `Cargo.toml` where the version is pinned then

atuchin-m: added a comment.

PR #530, opened 2025-10-17
0.11.x merge to master

atuchin-m:
The commits were rebased in a new branch `0.11.x-merge` (to preserve the original commits).


antonok-edm (approved):
thanks!

src/blocker.rs
```diff
@@ -66,26 +67,16 @@ static NO_TAGS: Lazy<HashSet<String>> = Lazy::new(HashSet::new);
...
-
     // Enabled tags are not serialized - when deserializing, tags of the existing
     // instance (the one we are recreating lists into) are maintained
     pub(crate) tags_enabled: HashSet<String>,
-    pub(crate) tagged_filters_all: NetworkFilterList,
-
-    pub(crate) enable_optimizations: bool,
-
     // Not serialized
-    #[cfg(feature = "unsync-regex-caching")]
```
bridiver, line 85: renaming should not be combined with other changes

atuchin-m: It was a request from @antonok-edm 
https://github.com/brave/adblock-rust/pull/489#discussion_r2164994795

bridiver: and that request should have been rejected here and done in a separate PR

atuchin-m: For b-c it doesn't change things, because we would get it in adblock 0.11.0 anyway.
Or you suggest to make a dedicated adblock-rust release with renaming only?

bridiver: This change broke brave-core because it kept `unsync-regex-caching` so `single-thread` was not applied. The same principles that apply to brave-core apply here as well so it should not have been part of this PR and per semver rules it should have been a dedicated adblock-rust release because it's not backwards compatible so should have been a major version bump. At the very least it should have been a separate PR so it didn't get lost among a bunch of other changes.

PR #531, opened 2025-10-20
Pre release 0.11.x fixes

atuchin-m:
The PR:
* fixes `decode_script_with_permission` issues with some `PermissionMask`;
* exposes `ResourceImpl` and `EngineDebugInfo`.

P.S. the perf issue with `blocker_new/brave-list-deserialize` isn't relevant (checked locally). It's likely a fast runner was used for reference.

### DIFF
--- a/src/cosmetic_filter_utils.rs
+++ b/src/cosmetic_filter_utils.rs
@@ -84,19 +84,22 @@ impl SpecificFilterType {
     }
 }
 
-/// Encodes permission bits in the last 2 ascii bytes of a script string
+/// Encodes permission bits in the last 2 ascii chars of a script string
 /// Returns the script with permission appended
 pub(crate) fn encode_script_with_permission(
     mut script: String,
     permission: PermissionMask,
 ) -> String {
-    let permission_string = format!("{:02x}", permission.to_bits());
-    debug_assert_eq!(permission_string.len(), 2);
-    script.push_str(&permission_string);
+    const HEX_CHARS: &[u8; 16] = b"0123456789abcdef";
+    let high = (permission.to_bits() >> 4) as usize;
+    let low = (permission.to_bits() & 0x0f) as usize;
+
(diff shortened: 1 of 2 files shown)


src/cosmetic_filter_utils.rs
```diff
@@ -94,16 +94,41 @@ pub(crate) fn encode_script_with_permission(
     script
 }
 
-/// Decodes permission bits from the last byte of a script string
+/// Decodes permission bits from the last char of a script string
 /// Returns (permission, script) tuple
 pub(crate) fn decode_script_with_permission(encoded_script: &str) -> (PermissionMask, &str) {
```
antonok-edm, line 104: good catch. Can we change encoded script + permissions to be `&[u8]` in all instances? Should make it harder to misuse, especially considering that Rust APIs widely assume a `&str` is valid Unicode data.

src/lib.rs
```diff
@@ -42,6 +42,10 @@ pub use engine::Engine;
 #[doc(inline)]
 pub use lists::FilterSet;
 
+#[cfg(feature = "debug-info")]
+#[doc(inline)]
+pub use engine::EngineDebugInfo;
```
antonok-edm: I'd prefer to expose `pub mod engine` rather than adding this as a top-level export, just to keep the documentation cleaner. It's too advanced of a use-case for first-page docs placement.

We can still keep the `pub use engine::Engine` re-export since that is almost always going to be used.

src/cosmetic_filter_utils.rs
atuchin-m: Unfortunately, that's problematic: it hits the flatbuffer format limitations. It doesn't support arrays of arrays, so we have to wrap it in a new table. That will consume extra memory and makes the code less clear.

src/lib.rs
atuchin-m: Done.

antonok-edm (approved):

PR #540, opened 2025-10-30
Optimize network filter get tokens

atuchin-m:
before:
get_token_before

after:
get_token_after

### DIFF
--- a/src/filters/fb_network_builder.rs
+++ b/src/filters/fb_network_builder.rs
@@ -138,7 +138,7 @@ impl<'a> FlatSerialize<'a, EngineFlatBuilder<'a>> for NetworkFilterListBuilder {
             .filters
             .into_iter()
             .map(|filter| {
-                let tokens = filter.get_tokens();
+                let tokens = filter.get_tokens_optimized();
                 (filter, tokens)
             })
             .collect();
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -882,7 +882,18 @@ impl NetworkFilter {
         )
     }
 
-    pub fn get_tokens(&self) -> FilterTokens {
+    #[deprecated(since = "0.11.1", note = "use get_tokens_optimized instead")]
+    pub fn get_tokens(&self) -> Vec<Vec<Hash>> {
+        match self.get_tokens_optimized() {
+            FilterTokens::OptDomains(domains) => {
+                domains.into_iter().map(|domain| vec![domain]).collect()
+            }
+            FilterTokens::Other(tokens) => vec![tokens],
+            FilterTokens::Empty => vec![],
+        }
+    }
+
+    pub fn get_tokens_optimized(&self) -> FilterTokens {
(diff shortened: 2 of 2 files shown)


src/network_filter_list.rs
```diff
@@ -155,16 +155,25 @@ impl NetworkFilterList<'_> {
...
+                    *tokens_histogram.entry(to_short_hash(*t)).or_insert(0) += 1;
+                    number_of_tokens += 1;
+                }
+            }
+            FilterTokens::Other(tokens) => {
+                for t in tokens {
+                    *tokens_histogram.entry(to_short_hash(*t)).or_insert(0) += 1;
+                    number_of_tokens += 1;
+                }
             }
```
antonok-edm: could be collapsed into `FilterTokens::OptDomains(tokens) | FilterTokens::Other(tokens) => {/* ... */}`

src/filters/fb_network_builder.rs
```diff
@@ -156,33 +156,55 @@ impl<'a> FlatSerialize<'a, EngineFlatBuilder<'a>> for NetworkFilterListBuilder {
...
+                                    min_count = 0;
+                                    best_token = token
+                                }
+                                Some(&count) if count < min_count => {
+                                    min_count = count;
+                                    best_token = token
+                                }
+                                _ => {}
+                            }
+                        }
```
antonok-edm, line 197: let's make the `match` return `best_token` as `Option<Hash>`, so the part where it's used doesn't need to be duplicated

src/filters/fb_network_builder.rs
```diff
@@ -156,33 +156,55 @@ impl<'a> FlatSerialize<'a, EngineFlatBuilder<'a>> for NetworkFilterListBuilder {
...
-                    for token in tokens {
-                        let token = to_short_hash(token);
-                        match tokens_histogram.get(&token) {
-                            None => {
-                                min_count = 0;
-                                best_token = token
-                            }
-                            Some(&count) if count < min_count => {
-                                min_count = count;
-                                best_token = token
```
antonok-edm, line 171: does the speedup come from removing this `tokens_histogram` check for token groups of length 1, or from removing the extra `Vec` allocations by using `FilterTokens` instead? The former could be done without a breaking change; just trying to understand the options we have

atuchin-m: The main profit is originated from removing extra allocations.

src/filters/fb_network_builder.rs
atuchin-m: I added `store_filter`, because for `FilterTokens::OptDomains` we added multiple best tokens.

src/filters/fb_network_builder.rs
atuchin-m: I've returned the original method to maintain API compatibility.

src/network_filter_list.rs
atuchin-m: Done

PR #543, opened 2025-10-31
Build cosmetic filters first to optimize peak memory usage

atuchin-m:
For https://github.com/brave/brave-browser/issues/50463


atuchin-m:
It's just -4MB of peak memory, too small optimization to merge.

PR #553, opened 2025-11-05
[perf] use arrayvec to save memory & CPU

atuchin-m:
The PR reduces the amount of allocation by introducing a new struct `StackVector` that allocates memory outside on a stack instead of a heap. 
* the number of allocations is reduced (`-19%, memory-usage/brave-list-initial/alloc-count`);
* building time is improved (`-11-15%, memory-usage/brave-list-initial`);
* the tests expectations were updated, because before one case hits the token limit.

Also, I tried to use `MaybeUinit` + `unsafe` to avoid zeroing the memory, but it turns out it doesn't give any profit now, so let's stash that idea for future usage.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -29,6 +29,8 @@ fn is_allowed_filter(ch: char) -> bool {
     ch.is_alphanumeric() || ch == '%'
 }
 
+/// A fixed-size array-like vector of hashes with maximum capacity of 256.
+/// Used instread of Vec<Hash> to avoid heap allocations.
 pub type TokensBuffer = ArrayVec<Hash, 256>;
 
 fn fast_tokenizer_no_regex(


src/utils.rs
```diff
@@ -27,25 +29,23 @@ fn is_allowed_filter(ch: char) -> bool {
     ch.is_alphanumeric() || ch == '%'
 }
 
-pub(crate) const TOKENS_BUFFER_SIZE: usize = 128;
-pub(crate) const TOKENS_BUFFER_RESERVED: usize = 1;
-const TOKENS_MAX: usize = TOKENS_BUFFER_SIZE - TOKENS_BUFFER_RESERVED;
+pub type TokensBuffer = StackVector<Hash, 256>;
```
boocmp: Why 256, if there was 128 (127 + 1 for zero slot) ?

tests/matching.rs
```diff
@@ -199,5 +199,5 @@ fn check_rule_matching_browserlike() {
     let (blocked, passes) = bench_rule_matching_browserlike(&engine, &requests);
     let msg = "The number of blocked/passed requests has changed. ".to_string()
         + "If this is expected, update the expected values in the test.";
-    assert_eq!((blocked, passes), (106860, 136085), "{msg}");
+    assert_eq!((blocked, passes), (106861, 136084), "{msg}");
```
boocmp, line 202: Why is this expected ?

src/flatbuffers/unsafe_tools.rs
```diff
@@ -97,3 +97,57 @@ impl VerifiedFlatbufferMemory {
         &self.raw_data[self.start..]
     }
 }
+
+/// A stack-allocated vector that uses [T; MAX_SIZE] with Default initialization.
+/// All elements are initialized to T::default(), and we track the logical size separately.
+/// Note: a future impl can switch to using MaybeUninit with unsafe code for better efficiency.
+pub struct StackVector<T, const MAX_SIZE: usize> {
```
boocmp: It seems to me, this shouldn't be in `unsafe_tools`, because it looks totally safe

src/flatbuffers/unsafe_tools.rs
```diff
@@ -97,3 +97,57 @@ impl VerifiedFlatbufferMemory {
...
+    fn default() -> Self {
+        Self {
+            data: [T::default(); MAX_SIZE],
+            size: 0,
+        }
+    }
+}
+
+impl<T: Default, const MAX_SIZE: usize> StackVector<T, MAX_SIZE> {
+    pub fn push(&mut self, value: T) -> bool {
```
boocmp: Do you mind to panic instead of returning bool ?

src/request.rs
```diff
@@ -239,11 +239,11 @@ impl Request {
 }
 
 fn calculate_tokens(url_lower_cased: &str) -> Vec<utils::Hash> {
```
boocmp, line 241: Is it possible to use TokensBuffer directly w/o converting it into the Vec<..> ?

src/flatbuffers/unsafe_tools.rs
atuchin-m: it's totally safe now, but the future impl should became unsafe. We can move it, if you want to.

src/flatbuffers/unsafe_tools.rs
atuchin-m: I believe we shouldn't panic by default. 
If a caller need, it can use `if (!push(..)) { panic!(...); }`

src/request.rs
atuchin-m: `Request` (usually allocated on a heap) stores it as `pub(crate) request_tokens: Vec<utils::Hash>`.
If we change that, the Request size is increased.
Sounds like it's not we want to.

src/utils.rs
atuchin-m: We used different limits:
`pub(crate) const TOKENS_BUFFER_SIZE: usize = 200;`
Now we use 256 everywhere.
So it's must be at least 200. 256 is just to make it a machine-friendly, it doesn't effect the perf.
I can revert it to 200.

tests/matching.rs
atuchin-m: I mentioned in the PR description: we hits the token limits for one rule.
If you take master and increase the limits, you will get the same behavior.

src/request.rs
boocmp: there is only one request for check per matching call, so I don't care about its size :)

src/flatbuffers/unsafe_tools.rs
boocmp: Ah, I see.

src/request.rs
boocmp: Discussed in DM, it doesn't affect the prod perf because there is only one request, and replacing Vec doesn't simplify the code, so lets leave it as is

tests/matching.rs
boocmp: nvm, the reason is in the descr

src/flatbuffers/unsafe_tools.rs
boocmp: Do you mind to rename it to InlineVec or ArrayVec?

atuchin-m: Renamed to `ArrayVec`

src/utils.rs
```diff
@@ -6,6 +6,60 @@ use seahash::hash;
 #[cfg(target_pointer_width = "32")]
 use seahash::reference::hash;
 
+/// A stack-allocated vector that uses [T; MAX_SIZE] with Default initialization.
+/// All elements are initialized to T::default(), and we track the logical size separately.
+/// Note: a future impl can switch to using MaybeUninit with unsafe code for better efficiency.
+pub struct ArrayVec<T, const MAX_SIZE: usize> {
+    data: [T; MAX_SIZE],
+    size: usize,
+}
```
antonok-edm: Let's use the existing [`arrayvec` crate](https://crates.io/crates/arrayvec) for this. That implementation has already been optimized and battle-tested across the Rust ecosystem (along with the similar [`ArrayString`](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayString.html)) for a very long time now.

`arrayvec 0.7.6` is already in the dependency tree for brave-core.

atuchin-m: Nice catch!
I though we don't use `arrayvec` in b-c, so adding extra deps is undesirable.
Done.

src/utils.rs
```diff
@@ -27,25 +29,23 @@ fn is_allowed_filter(ch: char) -> bool {
     ch.is_alphanumeric() || ch == '%'
 }
 
-pub(crate) const TOKENS_BUFFER_SIZE: usize = 128;
-pub(crate) const TOKENS_BUFFER_RESERVED: usize = 1;
-const TOKENS_MAX: usize = TOKENS_BUFFER_SIZE - TOKENS_BUFFER_RESERVED;
+pub type TokensBuffer = ArrayVec<Hash, 256>;
```
antonok-edm, line 34: doc comment would be useful here

src/filters/network.rs
```diff
@@ -312,10 +310,10 @@ pub enum FilterPart {
 }
 
 #[derive(Debug, PartialEq)]
-pub enum FilterTokens {
+pub enum FilterTokens<'a> {
```
antonok-edm, line 313: this is now also a breaking change :smiling_face_with_tear:

atuchin-m: hm, the same issue with `get_tokens_optimized`.
Not sure that it makes sense to introduce `get_tokens_optimized2`, it sounds like it's about time to release 0.12.0

src/utils.rs
atuchin-m: added.

PR #554, opened 2025-11-06
[perf] tokenize some regexp patterns

atuchin-m:
The PR tokenizes some common regex filters like `||dlscord.*$all` (with token `dlscord`). This reduces the number of filers that always checked (stored with token = `0`) that improves the filter matching performance.

`-13%` on my machine:
```
rule-match-browserlike/brave-list
master:  time:    [1.5573 s 1.5734 s 1.5927 s]
this PR:  time:   [1.3547 s 1.3618 s 1.3694 s]
```

### DIFF
--- a/tests/unit/engine.rs
+++ b/tests/unit/engine.rs
@@ -237,9 +237,9 @@ mod tests {
             );
         }
         let expected_hash: u64 = if cfg!(feature = "css-validation") {
-            9439492009815519037
+            15545091389304905433
         } else {
-            14803842039735157685
+            543362704487480180
         };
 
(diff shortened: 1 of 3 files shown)


tests/unit/engine.rs
```diff
@@ -237,9 +237,9 @@ mod tests {
             );
         }
         let expected_hash: u64 = if cfg!(feature = "css-validation") {
-            9439492009815519037
+            15545091389304905433
```
atuchin-m, line 240: We don't need to update `ADBLOCK_RUST_DAT_VERSION`, because the matching logic isn't changed. Tt's only a optimization how we store rules. The current .dat files can be used as is.

PR #560, opened 2025-11-11
Don't use old method in tests

atuchin-m:
The PR removes `impl NetworkMatchable for NetworkFilter` and port the related tests to the flatbuffer impl (that we actually use in production)

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -135,6 +135,8 @@ impl Blocker {
         self.get_list(NetworkFilterListId::TaggedFiltersAll)
     }
 
+    /// Borrow mutable reference to the regex manager for the ['Blocker`].
+    /// Only one caller can borrow the regex manager at a time.
     pub(crate) fn borrow_regex_manager(&self) -> RegexManagerRef<'_> {
         #[cfg(feature = "single-thread")]
         #[allow(unused_mut)]
@@ -155,7 +157,7 @@ impl Blocker {
             .is_some()
     }
 
-    #[doc(hidden)]
+    #[cfg(test)]
     pub(crate) fn check_exceptions(&self, request: &Request) -> bool {
         let mut regex_manager = self.borrow_regex_manager();
         self.exceptions()
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -139,7 +139,7 @@ impl Engine {
         self.blocker.check(request, &self.resources)
     }
 
-    #[doc(hidden)]
+    #[cfg(test)]
     pub(crate) fn check_network_request_exceptions(&self, request: &Request) -> bool {
         self.blocker.check_exceptions(request)
     }
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -970,8 +970,8 @@ impl NetworkFilter {
         }
     }
 
-    #[doc(hidden)]
-    pub fn matches_test(&self, request: &request::Request) -> bool {
+    #[cfg(test)]
+    pub(crate) fn matches_test(&self, request: &request::Request) -> bool {
         let filter_set = crate::FilterSet::new_with_rules(vec![self.clone()], vec![], true);
         let engine = crate::Engine::from_filter_set(filter_set, true);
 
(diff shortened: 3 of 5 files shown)


tests/unit/filters/network_matchers.rs
```diff
@@ -659,22 +688,34 @@ mod match_tests {
...
-            let mut regex_manager = RegexManager::default();
-            assert!(regex_manager.get_compiled_regex_count() == 0);
+            assert_eq!(
+                engine
+                    .get_debug_info()
+                    .regex_debug_info
+                    .compiled_regex_count,
+                0
+            );
+            // Regex can't be compiled, so no match.
```
atuchin-m, line 707: the test case added here (already in the disabled state):
https://github.com/brave/adblock-rust/commit/40b7815585685feed8be806bea0f1246f7b0cbd4

regex crate can't parse such regex, so I changed the test to verify it doesn't match

src/blocker.rs
```diff
@@ -130,31 +135,34 @@ impl Blocker {
...
-
     pub fn check_generic_hide(&self, hostname_request: &Request) -> bool {
         let mut regex_manager = self.borrow_regex_manager();
         self.generic_hide()
             .check(hostname_request, &HashSet::new(), &mut regex_manager)
             .is_some()
     }
 
+    #[doc(hidden)]
+    pub(crate) fn check_exceptions(&self, request: &Request) -> bool {
```
antonok-edm: no need to include `#[doc(hidden)]` on `pub(crate)` items

src/filters/network.rs
```diff
@@ -969,6 +969,18 @@ impl NetworkFilter {
             FilterTokens::Other(t)
         }
     }
+
+    #[doc(hidden)]
+    pub fn matches_test(&self, request: &request::Request) -> bool {
```
antonok-edm: `#[cfg(test)]`?

tests/unit/regex_manager.rs
```diff
@@ -25,45 +24,65 @@ mod tests {
 
     #[test]
     fn simple_match() {
-        let mut regex_manager = RegexManager::default();
-        regex_manager.update_time();
+        let engine = make_engine("||geo*.hltv.org^");
+        engine.borrow_regex_manager();
```
antonok-edm: comment would also be useful here to explain why the extra borrows are necessary

src/engine.rs
```diff
@@ -266,6 +271,11 @@ impl Engine {
         self.blocker.set_regex_discard_policy(new_discard_policy);
     }
 
+    #[cfg(test)]
+    pub fn borrow_regex_manager(&self) -> crate::blocker::RegexManagerRef<'_> {
```
antonok-edm, line 275: comment describing why this is necessary would be helpful

(I can tell from the diffs below, but later it won't be as obvious)

src/filters/network.rs
```diff
@@ -993,30 +1005,6 @@ pub trait NetworkMatchable {
     fn matches_test(&self, request: &request::Request) -> bool;
 }
 
-impl NetworkMatchable for NetworkFilter {
-    fn matches(&self, request: &request::Request, regex_manager: &mut RegexManager) -> bool {
```
antonok-edm, line 997: fwiw, this is a breaking API change and will require another bump up to `0.12.x`

src/engine.rs
atuchin-m: It's just a test wrapper for `blocker.borrow_regex_manager`, so I added a doc comment to `borrow_regex_manager()` there.

tests/unit/regex_manager.rs
atuchin-m: Actually, we don't need this with thread-local mock time. Removed.

src/filters/network.rs
atuchin-m: `matches_test` was used `tests/matching.rs` for exception checking.
I've adjusted the tests to use directly `engine.check_network_request` and added `#[cfg(test)]`

src/blocker.rs
atuchin-m: fixed, thanks

PR #574, opened 2025-11-20
pre-0.12.x => master

atuchin-m:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "adblock"
-version = "0.11.1"
+version = "0.12.0"
```
antonok-edm, line 3: we need a version bump in a couple of other places too. My workflow is this:

```
1. Update versions in `Cargo.toml`, `package.json`, `js/Cargo.toml`
2. Trigger `Cargo.lock` updates (i.e. `cargo update --workspace` or any kind of `cargo build` command)
3. Trigger `package-lock.json` updates (i.e. `npm update adblock-rs` or any kind of `npm install` command)
```

src/filters/network.rs
```diff
@@ -885,17 +883,21 @@ impl NetworkFilter {
...
+        match self.get_tokens_optimized(&mut tokens_buffer) {
             FilterTokens::OptDomains(domains) => {
-                domains.into_iter().map(|domain| vec![domain]).collect()
+                domains.iter().map(|domain| vec![*domain]).collect()
             }
-            FilterTokens::Other(tokens) => vec![tokens],
+            FilterTokens::Other(tokens) => vec![tokens.to_vec()],
             FilterTokens::Empty => vec![],
         }
     }
```
antonok-edm: since we're making a breaking version change, shall we remove this one also?

we can rename `get_tokens_optimized` back to `get_tokens` in the process

antonok-edm: also, I am open to making these `pub(crate)` to prevent needing similar breaking changes in the future

src/filters/network.rs
```diff
@@ -993,30 +1005,6 @@ pub trait NetworkMatchable {
     fn matches_test(&self, request: &request::Request) -> bool;
 }
 
-impl NetworkMatchable for NetworkFilter {
```
antonok-edm, line 996: I think we can also make `NetworkMatchable` into a `pub(crate)` trait since it's only implemented for `FlatNetworkFilter` now

atuchin-m: Done. I also removed unused `matches_test`.
Maybe it makes sense to completely remove trait `NetworkMatchable`, but since it non-public, we can do it later.

src/filters/network.rs
atuchin-m: Done. I also made `pub(crate)` related `TokensBuffer` and `FilterTokens`

PR #584, opened 2026-01-10
Fix #479: Smart www-stripping for hostname anchor filters

### LINKED ISSUE #479: Blocking the "www" website with any TLD will block almost every websites with that TLD
Hi.

If we block the `www` website of any TLD,
for example:
```
||www.com^
```
it will block almost every `.com` websites.

blocked:
https://example.com
https://github.com
https://youtube.com
https://google.com
https://chatgpt.com
https://copilot.microsoft.com
... etc (almost everything)

some very few websites seems to be immune (there are probably other exceptions):
https://brave.com
https://news.ycombinator.com

Didn't find a pattern about why these 2 above websites are not blocked.

The rule works as expected in uBlock Origin.

example of an affected adblock list:
The ["multi ultimate" list](https://github.com/hagezi/dns-blocklists/tree/main?tab=readme-ov-file#closed_book-multi-ultimate---aggressive-protection-) from @hagezi has `||www.be^` (malware website) in it, and therefore, every `.be` websites are blocked.

I didn't find any opened issue reporting this problem. Neither here or on `brave/brave-browser`.

R0h1tAnand:
## Problem

Hostname anchor filters like `||www.com^` were having the `www.` prefix unconditionally stripped during filter parsing (line 757 in `src/filters/network.rs`). This caused the filter hostname to become `"com"` instead of `"www.com"`.

When matching requests, `"com"` would match as a suffix of any `.com` hostname, causing virtually all `.com` domains to be incorrectly blocked. The same issue affected other TLDs: `||www.be^` would block all `.be` websites.

**Affected Filter Lists:**
- Hagezi's "Multi Ultimate" list contains `||www.be^` (a malware domain), causing all `.be` websites to be blocked.

**Expected Behavior (uBlock Origin):**
- `||www.com^` should only block `www.com` and its subdomains (e.g., `sub.www.com`)
- `||www.com^` should NOT block `example.com`, `github.com`, etc.

## Solution

Implemented smart www-stripping logic that only removes the `www.` prefix when it represents a subdomain, not when it's the actual domain name:

### Before Fix:
- `||www.com^` ??? hostname: `"com"` ??? blocks ALL .com domains ???
- `||www.youjizz.com^` ??? hostname: `"youjizz.com"` ??? works correctly ???

### After Fix:
- `||www.com^` ??? hostname: `"www.com"` ??? blocks only www.com and subdomains ???
- `||www.youjizz.com^` ??? hostname: `"youjizz.com"` ??? still works correctly ???

### Logic:
The fix checks if stripping `www.` would leave a valid domain with at least one dot:
- If `after_www.contains('.')`: strip www. (it's a subdomain prefix)
- Otherwise: preserve www. (it's the actual domain)

## Changes

- **src/filters/network.rs**: Implemented smart www-stripping logic in hostname parsing
- **src/filters/network_matchers.rs**: Code restructuring for clarity (no functional change to matching logic)
- **tests/unit/engine.rs**: Updated serialization hash to reflect the fix

## Testing

All existing tests pass, including:
- `hostname_regex_filter_works`: Wildcard hostname filters still work
- `is_anchored_by_hostname_works`: Hostname matching logic preserved
- `check_pattern_hostname_anchor_filter_works`: All existing hostname anchor tests pass

### Manual Testing Verified:
- ??? `||www.com^` now only blocks `www.com` (not `example.com`, `github.com`, etc.)
- ??? `||www.be^` now only blocks `www.be` (not all `.be` domains)
- ??? `||www.youjizz.com^` correctly blocks `cdne-pics.youjizz.com` (existing behavior preserved)

## Breaking Changes

The serialization format hash has changed. Filter lists using `||www.[tld]^` filters will now work correctly instead of over-blocking.

Fixes #479

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -753,10 +753,24 @@ impl NetworkFilter {
 
         let hostname_decoded = hostname
             .map(|host| {
+                // Strip "www." prefix only if it's a subdomain (not the whole domain)
+                // ||www.com^ should NOT be stripped (www.com is the actual domain)
+                // ||www.youjizz.com^ SHOULD be stripped to youjizz.com (www is a subdomain)
                 let hostname_normalised = if mask.contains(NetworkFilterMask::IS_HOSTNAME_ANCHOR) {
-                    host.trim_start_matches("www.")
+                    // Check if stripping www. would leave a valid domain with at least one dot
+                    if let Some(after_www) = host.strip_prefix("www.") {
+                        if after_www.contains('.') {
+                            // www. is a subdomain prefix, strip it
+                            after_www.to_string()
+                        } else {
+                            // www. is the whole domain (e.g., www.com), don't strip
+                            host
+                        }
+                    } else {
+                        host
+                    }
                 } else {
-                    &host
+                    host
                 };
 
                 let lowercase = hostname_normalised.to_lowercase();
--- a/src/filters/network_matchers.rs
+++ b/src/filters/network_matchers.rs
@@ -41,35 +41,45 @@ fn is_anchored_by_hostname(
     } else if filter_hostname_len == hostname_len {
         // If they have the same len(), they should be equal
         filter_hostname == hostname
-    } else if let Some(match_index) = memmem::find(hostname.as_bytes(), filter_hostname.as_bytes())
-    {
-        if match_index == 0 {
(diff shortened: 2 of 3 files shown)


agadiffe:
This issue has been fixed in https://github.com/brave/adblock-rust/pull/618.
It can be closed.

PR #586, opened 2026-01-18
Fix: Add wildcard (*) support for custom filters

### LINKED ISSUE #582: Support for wildcards in Custom filters
*##+js(filter) // not working in brave shields , but working in ublock origin

please add support for wildcard it is important feature...

fresh3nough:
## Summary
Fixes #582 

This PR adds support for wildcard (`*`) syntax in cosmetic filters, particularly for scriptlet injection rules like `*##+js(filter)`. This brings adblock-rust's behavior in line with uBlock Origin and enables filter list authors to apply scriptlets globally across all domains.

## Problem
Previously, filters using the wildcard syntax `*##+js(scriptlet)` would parse successfully but fail to match any domains. The parser treated `*` as a literal hostname rather than recognizing it as a wildcard meaning "all domains". This caused the filter to have hostname constraints that would never match.

## Solution
The fix is minimal and elegant - we simply skip the `*` wildcard during domain parsing in the `locations_before_sharp()` function

### DIFF
--- a/tests/unit/engine.rs
+++ b/tests/unit/engine.rs
@@ -237,9 +237,9 @@ mod tests {
             );
         }
         let expected_hash: u64 = if cfg!(feature = "css-validation") {
-            7671576097086431151
+            9113803688770546221
         } else {
-            7220740598028920672
+            15343284834809453004
         };
 
         assert_eq!(hash(&data), expected_hash, "{HASH_MISMATCH_MSG}");


fresh3nough:
Fixed all the issues:

1. Moved the validation logic to properly reject wildcard script injections - they now correctly fail since generic script injections aren't allowed
2. Removed the nonsensical `*,example.com` test case 
3. Added runtime matching tests that verify wildcards actually work at runtime, not just during parsing

The wildcard filters now work as generic rules - `*##selector` applies to all domains just like `##selector` does.

PR #588, opened 2026-01-18
Enable css-validation feature by default to fix :has-text() selector

### LINKED ISSUE #537: has-text selector does not work
It seems you do have support for the `has-text` selector?

https://github.com/brave/adblock-rust/blob/1bf068dbe186d89b3720c8c1ae59525ffde660fe/tests/unit/filters/cosmetic.rs#L714-L720

But this does not seem to work in Brave? Anything with `:has-text(...)` in it does not seem to parse correctly?

It's incredibly hard to tell what's going on - in the "block element" dialog, there is no indication of whether or not your selector is even valid. You can't tell if your selector has a syntax error, or it just doesn't match any elements. I don't understand how people develop or test filters for this ad blocker? Is this really the best/only tool we have for this?

(I want to block "Reels" on facebook, but there are literally no stable selectors to be had anywhere on the page - every single ID and class name are generated at build-time and will change with their next deployment. The only reasonably stable selector would be a text selector. I know this would have performance implications.)

fresh3nough:
The :has-text() selector and other procedural operators were not working because the css-validation feature was not enabled by default. Without this feature, all selectors are treated as plain CSS instead of being properly parsed to extract procedural operators.

This fix enables css-validation as a default feature, allowing :has-text() and other procedural selectors to work correctly.

Fixes #537

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,8 +92,8 @@ harness = false
 
 [features]
 # If disabling default features, consider explicitly re-enabling the
-# "embedded-domain-resolver" feature.
-default = ["embedded-domain-resolver", "full-regex-handling", "single-thread"]
+# "embedded-domain-resolver" and "css-validation" features.
+default = ["embedded-domain-resolver", "full-regex-handling", "single-thread", "css-validation"]
 full-regex-handling = []
 single-thread = [] # disables `Send` and `Sync` on `Engine`.
 debug-info = []


mindplay-dk:
Why was this closed?

Without any comments?

fresh3nough:
my apologies, i was cleaning up, if you would like me to reopen it, I gladly will

mindplay-dk:
@fresh3nough yes, please? I run into many things that can't be blocked without this selector - there is so much client-side UI now where elements are div soup and every CSS class-name is machine-generated. 😊

PR #594, opened 2026-02-05
Export ADBLOCK_RUST_DAT_VERSION constant

gutchin1:


gutchin1:
> Can you say more about your use case?

I am emedding the crate in my project, and since the serialization format gets updated, I need a way for other components to be aware of this change and provide the serialized input accordingly

antonok-edm:
> I am emedding the crate in my project, and since the serialization format gets updated, I need a way for other components to be aware of this change and provide the serialized input accordingly

I think you might be misinterpreting the purpose of the serialized format... it's intended to be used as a local cache, so that it's faster to load previously-parsed filters (e.g. at application startup). It's not intended as a full replacement for filter lists.

If the format version number has incremented, [`deserialize`](https://docs.rs/adblock/latest/adblock/struct.Engine.html#method.deserialize) will immediately exit with a `VersionMismatch` error. So the worst case is that cheap check to discover that the format is outdated followed by reparsing the filters (which still shouldn't be extremely slow either).

antonok-edm:
closing due to no response

PR #600, opened 2026-02-27
test: Add JS binding tests

mschfh:


js/test/bindings.test.mjs
```diff
@@ -0,0 +1,648 @@
...
+describe('Engine.check — exception rules', () => {
+    it('exception rule prevents blocking, populates exception field', () => {
+        const fs = new FilterSet(true);
+        fs.addFilters(['||ads.example.com^', '@@||ads.example.com^$domain=publisher.com']);
+        const engine = new Engine(fs, true);
+        const result = engine.check(
+            'https://ads.example.com/tracker.js', 'https://publisher.com', 'script', true,
+        );
+        assert.equal(result.matched, false);
+        assert.equal(typeof result.exception, 'string');
```
antonok-edm, line 205: would be good to confirm the rule string matches the original rule added to the list (same for other tests where applicable via `filter` field)

js/test/bindings.test.mjs
```diff
@@ -0,0 +1,648 @@
...
+        engine.clearTags();
+        assert.equal(engine.tagExists('stuff'), false);
+        assert.equal(engine.tagExists('brian'), false);
+        assert.equal(engine.check('https://example.com/adv', 'https://example.com', 'other'), false);
+    });
+});
+
+// NOTE: Engine.addResource is listed in js/index.js forwardClassMethods but
+// Engine_addResource is NOT exported in register_module! (js/src/lib.rs).
+// It will throw at runtime. Skipping test until the export is added.
```
antonok-edm: Missed removing that one earlier, let's remove it from `js/index.js` as well as this comment

js/test/bindings.test.mjs
atuchin-m, line 1: Thanks for adding these tests.
I would say that the primary tests should be in Rust and here we likely don't want to duplicate tests.

I wonder can we reduce some logic check check in favor of verifing that API works = proxy to the matching rust method).

I.e. `$third-party and $1p modifiers` looks excessive.

antonok-edm: > I would say that the primary tests should be in Rust and here we likely don't want to duplicate tests.

I agree that primary tests should be in Rust, but I don't think there's a compelling reason to remove/reduce JS tests - they can help catch regressions or API changes and have very little cost if delegated to agentic coding assistants.

> I wonder can we reduce some logic check check in favor of verifing that API works = proxy to the matching rust method).

Definitely agree with this one, but I don't see any easy way of doing it at the moment :smiling_face_with_tear:

PR #604, opened 2026-03-08
refactor: Migrate Neon JS bindings to NAPI

mschfh:


js/src/lib.rs
```diff
@@ -4,421 +4,273 @@ use adblock::lists::{
...
+    inner: RefCell<FilterSetInternal>,
+}
 
-        match serde_json::from_str(&js_string.value(cx)) {
-            Ok(v) => Ok(v),
-            Err(e) => cx.throw_error(e.to_string())?,
+// Safety: the `single-thread` feature on adblock makes FilterSetInternal !Send,
+// but NAPI-RS classes require Send. The JS runtime is single-threaded so this is
+// safe in practice.
+unsafe impl Send for JsFilterSet {}
```
antonok-edm, line 26: this looks sketchy, why do NAPI-RS classes require Send? We should understand and document that before blindly adding `Send` here.

js/src/lib.rs
```diff
@@ -4,421 +4,273 @@ use adblock::lists::{
...
+    pub fn add_filter(&self, filter: String, opts: Option<serde_json::Value>) -> Result<bool> {
+        let parse_opts: ParseOptions = match opts {
+            Some(v) => serde_json::from_value(v).map_err(|e| Error::from_reason(e.to_string()))?,
+            None => ParseOptions::default(),
+        };
+        Ok(self
+            .inner
+            .borrow_mut()
+            .add_filter(&filter, parse_opts)
+            .is_ok())
```
antonok-edm, line 61: > `Ok(_.is_ok())` 

This will silently swallow `FilterParseError`s, unlike the previous implementation

js/src/lib.rs
```diff
@@ -4,421 +4,273 @@ use adblock::lists::{
...
+            Ok((cb_rules, filters_used)) => {
+                let result = ContentBlockingConversionResult {
+                    content_blocking_rules: cb_rules,
+                    filters_used,
+                };
+                let val =
+                    serde_json::to_value(&result).map_err(|e| Error::from_reason(e.to_string()))?;
+                Ok(Either::A(val))
+            }
+            Err(_) => Ok(Either::B(())),
```
antonok-edm, line 76: API contract changed here again?

PR #608, opened 2026-03-16
Fix regex filter parsing when pattern contains $ end-anchor

### LINKED ISSUE #257: Support regex rules with an end-anchor (`$`)
Consider the following filter:
```adblock
/^https?:\/\/[a-z]{8,15}\.top\/[a-z]{4,}\.json$/$xhr,3p,match-case
```
There are two `$` characters in the filter. The first represents an anchor to the end of a URL, and the second separates the pattern part of the filter from the options.

`adblock-rust` currently treats the first `$` character as the separator, such that the pattern no longer ends with a `/` character. `match-case` is only supported for regex rules, so the rule is ultimately considered invalid.

0b0d3:
## Summary

Fixes #257

- Regex filters like `/pattern$/` use `$` as a regex end-of-string anchor, but the parser always treated the rightmost `$` as the options separator
- This caused filters like `/\.json$/` to fail with `UnrecognisedOption`, and filters like `/^https?:\/\/[a-z]{8,15}\.top\/[a-z]{4,}\.json$/$xhr,3p,match-case` to lose the `$` anchor from their pattern
- The fix tries each `$` from right to left, but only falls back to the next candidate when the pattern to the left looks like a regex (delimited by `/`). Non-regex filters still propagate option parse errors immediately

## Changes

- `src/filters/abstract_network.rs`: Replace single `find_char_reverse` with a loop that tries `$` candidates right-to-left, with regex-awareness to avoid false fallbacks
- `tests/unit/filters/network.rs`: Added 3 test cases covering regex with `$` anchor + options, regex with `$` anchor alone, and regex without `$` anchor
- `tests/unit/blocker.rs`: Updated easylist filter count (4 more filters now correctly parsed)
- `tests/unit/engine.rs`: Updated serialization hashes to reflect the new parsing behavior

## Test plan

- [x] All 262 existing tests pass (including with `css-validation` feature)
- [x] New test: regex filter with `$` end-anchor and options parses correctly
- [x] New test: regex filter with `$` end-anchor and no options parses correctly
- [x] New test: regex filter without `$` anchor and options still works
- [x] Non-regex filters with invalid options still return proper errors
- [x] Real-world filter lists (easylist, easyprivacy, ublock-unbreak) parse with correct counts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/filters/abstract_network.rs
+++ b/src/filters/abstract_network.rs
@@ -107,16 +107,68 @@ impl AbstractNetworkFilter {
             exception = true;
         }
 
-        let maybe_options_index: Option<usize> = find_char_reverse(b'$', line.as_bytes());
-
+        // Find the `$` that separates the pattern from filter options.
+        // Regex filters (delimited by `/`) may contain `$` as an end-of-string
+        // anchor. When a candidate `$` yields an options-parse error and the
+        // pattern to its left looks like a regex, try the next `$` to the left.
         let mut options = None;
-        if let Some(options_index) = maybe_options_index {
-            filter_index_end = options_index;
-
-            // slicing here is safe; the first byte after '$' will be a character boundary
-            let raw_options = &line[filter_index_end + 1..];
-
-            options = Some(parse_filter_options(raw_options)?);
+        {
+            let bytes = line.as_bytes();
+            let mut search_end = bytes.len();
+            let mut last_err = None;
+            loop {
+                let maybe_dollar = find_char_reverse(b'$', &bytes[..search_end]);
+                match maybe_dollar {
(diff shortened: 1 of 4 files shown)


ImpulseB23:
approach looks good, this is basically what i described in my comment above. logic is clean and the tests cover the key cases.

one small thing: the `pattern_before.ends_with("$/")` branch in the `looks_like_regex` check might be unreachable. when you're at a given `$`, `pattern_before` is everything to its left, so it can't end with `$/` unless there's a consecutive `$$` in the pattern. not a bug, just potentially dead code. might be worth dropping it or adding a test that hits it if you want to keep it.

otherwise lgtm

PR #638, opened 2026-05-11
forbid image-set in cosmetic filter styles

antonok-edm:
Yet another way to do network requests from CSS. https://hackerone.com/reports/3725810


antonok-edm:
`brave-list-deserialize` benchmark does not run any of the changed code, so it must be unrelated.

Issue #635, opened 2026-05-08
`NetworkFilterMask` is filled


antonok-edm:
The `NetworkFilterMask` bitflag struct is full; there's no room for additional information in it. This is hindering the adoption of new rule options and types.

Increasing the size of `NetworkFilterMask` from 32 to 64 bits is not a good solution because it will bloat the memory size of every parsed filter in the engine.

The optimal fix will involve reworking `NetworkFilter` and `NetworkFilterMask` to remove all the bits that are not actually checked at match time:
- `IS_CSP`
- `IS_IMPORTANT`
- `IS_REDIRECT`
- `IS_REMOVEPARAM`
- `GENERIC_HIDE`
- `BAD_FILTER`

These flags only determine which bin a network filter is placed into and not how they're matched at runtime, so they're wasted space in the in-memory representation of each filter bin. If we can return the information separately from the filter itself, we can discard it when adding it to the correct bin. So, parsing rules should return information about which bin a filter goes into alongside the filter itself.

Note all current bins:
- `csp`
- `importants`
- `redirects`
- `removeparam`
- `generic_hide`
- `tagged filters`
- `exceptions` (exception rules can appear in many bins, so we can't remove `IS_EXCEPTION`)
- `filters` (catch all for normal blocking filters)

antonok-edm:
this has been addressed by #641 and #642

PR #640, opened 2026-05-11
Refactor is_bad_filter()

atuchin-m:
For https://github.com/brave/adblock-rust/issues/635

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -21,6 +21,9 @@ use crate::utils::{self, Hash, TokensBuffer};
 static VALID_PARAM: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_\-]+$").unwrap());
 
 bitflags::bitflags! {
+  /// Features that are properties used to classify the filter, but not stored
+  /// in the flatbuffer serialized format. For that reason, not available
+  /// for FlatNetworkFilter (use NetworkFilterMask instead).
   #[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Default)]
   pub struct NetworkFilterFeaturesMask: u32 {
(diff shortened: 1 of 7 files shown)


src/filters/network.rs
```diff
@@ -100,7 +107,7 @@ bitflags::bitflags! {
         const THIRD_PARTY = 1 << 16;
         const FIRST_PARTY = 1 << 17;
         const IS_REDIRECT = 1 << 26;
-        const BAD_FILTER = 1 << 27;
+        const UNUSED_BAD_FILTER = 1 << 27;
```
antonok-edm: since it will be a breaking change anyways, we should remove this rather than renaming it

src/filters/network.rs
```diff
@@ -20,6 +20,13 @@ use crate::utils::{self, Hash, TokensBuffer};
 /// For now, only support `$removeparam` with simple alphanumeric/dash/underscore patterns.
 static VALID_PARAM: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9_\-]+$").unwrap());
 
+bitflags::bitflags! {
+  #[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Default)]
+  pub struct NetworkFilterFeaturesMask: u32 {
+    const BAD_FILTER = 1 << 0;
+  }
```
antonok-edm, line 30: let's also make sure the difference from `NetworkFilterMask` is documented

atuchin-m: done.

src/filters/network.rs
atuchin-m: removed.

atuchin-m:
Declined in favor of https://github.com/brave/adblock-rust/pull/641

PR #643, opened 2026-05-12
[perf, memory] Add single/multi_pattern to optimize memory usage (0.13.x)

atuchin-m:
The PR optimizes memory usage and building time by adding a simplified pattern storage.

Most filters (95%+) have `patterns.len() <= 1`. Despite that fact we always store `patterns` as a vector.
This implies storing an extra pointer (to the vector itself) + the vector length.

The PR adds an optional ability to store pattern as a single string eliminating the overhead above.

### DIFF
--- a/src/filters/fb_network.rs
+++ b/src/filters/fb_network.rs
@@ -8,24 +8,33 @@ use crate::regex_manager::RegexManager;
 use crate::request::Request;
 
 use crate::filters::flatbuffer_generated::fb;
+
 /// A list of string parts that can be matched against a URL.
-pub(crate) struct FlatPatterns<'a> {
-    patterns: Option<flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<&'a str>>>,
+pub(crate) enum FlatPatterns<'a> {
+    Empty,
+    Single(&'a str),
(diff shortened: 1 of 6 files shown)


src/filters/fb_network.rs
```diff
@@ -8,24 +8,33 @@ use crate::regex_manager::RegexManager;
...
 use crate::filters::flatbuffer_generated::fb;
+
 /// A list of string parts that can be matched against a URL.
-pub(crate) struct FlatPatterns<'a> {
-    patterns: Option<flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<&'a str>>>,
+pub(crate) enum FlatPatterns<'a> {
+    Empty,
+    Single(&'a str),
+    Multi(flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<&'a str>>),
 }
```
antonok-edm, line 21: would be nice to document the rationale for the optimization so an agent doesn't try to "fix" it back

antonok-edm (approved):

src/filters/fb_network.rs
antonok-edm: (I added it)

PR #652, opened 2026-05-26
New API for FilterSet & Engine

atuchin-m:
The PR introduces a new API to avoid working with `Vec<SomeFilter>`.
Instead FilterSet stores a set of annotated text sources.
This is necessary step to make streaming parsing work.

### DIFF
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -51,6 +51,31 @@ mod json_ffi {
     }
 }
 
+fn console_warn<'a, C: Context<'a>>(cx: &mut C, message: &str) -> NeonResult<()> {
+    let console: Handle<JsObject> = cx.global().get(cx, "console")?;
+    let warn: Handle<JsFunction> = console.get(cx, "warn")?;
+    let msg = JsString::new(cx, message).upcast::<JsValue>();
+    warn.call(cx, console, [msg])?;
+    Ok(())
+}
+
+/// Accepts a single filter list string, or a deprecated array of list strings (joined with `\n`).
+fn filter_rules_from_js<'a, C: Context<'a>>(
+    cx: &mut C,
+    input: Handle<JsValue>,
+) -> NeonResult<String> {
+    if input.downcast::<JsArray, _>(cx).is_ok() {
+        console_warn(
+            cx,
+            "FilterSet.addFilters: passing an array of strings is deprecated; \
+             pass a single newline-separated filter list string instead.",
+        )?;
+        let parts: Vec<String> = json_ffi::from_js(cx, input)?;
+        return Ok(parts.join("\n"));
+    }
+    json_ffi::from_js(cx, input)
+}
+
 #[derive(Serialize, Deserialize)]
 struct EngineOptions {
     pub optimize: Option<bool>,
@@ -91,7 +116,6 @@ fn create_filter_set(mut cx: FunctionContext) -> JsResult<JsBox<FilterSet>> {
 fn filter_set_add_filters(mut cx: FunctionContext) -> JsResult<JsValue> {
     let this = cx.argument::<JsBox<FilterSet>>(0)?;
 
-    // Take the first argument, which must be an array
     let rules_handle: Handle<JsValue> = cx.argument(1)?;
     // Second argument is optional parse options. All fields are optional. ParseOptions::default()
     // if unspecified.
@@ -100,7 +124,7 @@ fn filter_set_add_filters(mut cx: FunctionContext) -> JsResult<JsValue> {
         None => ParseOptions::default(),
     };
 
-    let rules: String = json_ffi::from_js(&mut cx, rules_handle)?;
+    let rules: String = filter_rules_from_js(&mut cx, rules_handle)?;
 
     let metadata = this.add_filters(rules, parse_opts);
 
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -149,9 +149,9 @@ fn engine_constructor(mut cx: FunctionContext) -> JsResult<JsBox<Engine>> {
                     config.optimize.unwrap_or(true)
                 }
             };
-            EngineInternal::from_filter_set(rules, optimize)
+            EngineInternal::new_with_filter_set(rules, optimize)
         }
-        None => EngineInternal::from_filter_set(rules, true),
+        None => EngineInternal::new_with_filter_set(rules, true),
     };
     Ok(cx.boxed(Engine(Mutex::new(engine_internal))))
 }
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -87,7 +87,7 @@ impl FilterSet {
     fn new(debug: bool) -> Self {
         Self(RefCell::new(FilterSetInternal::new(debug)))
     }
-    fn add_filters(&self, rules: String, opts: ParseOptions) {
+    fn add_filters(&self, rules: String, opts: ParseOptions) -> adblock::lists::FilterListMetadata {
         self.0.borrow_mut().add_filter_list(rules, opts)
     }
 
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -1,6 +1,4 @@
-use adblock::lists::{
-    FilterFormat, FilterListMetadata, FilterSet as FilterSetInternal, ParseOptions, RuleTypes,
-};
+use adblock::lists::{FilterFormat, FilterSet as FilterSetInternal, ParseOptions, RuleTypes};
 use adblock::resources::resource_assembler::assemble_web_accessible_resources;
 use adblock::resources::Resource;
 use adblock::Engine as EngineInternal;
@@ -64,16 +62,10 @@ impl FilterSet {
     fn new(debug: bool) -> Self {
         Self(RefCell::new(FilterSetInternal::new(debug)))
     }
-    fn add_filters(&self, rules: &[String], opts: ParseOptions) -> FilterListMetadata {
-        self.0.borrow_mut().add_filters(rules, opts)
+    fn add_filters(&self, rules: String, opts: ParseOptions) {
+        self.0.borrow_mut().add_filter_list(rules, opts)
     }
-    fn add_filter(
-        &self,
-        filter: &str,
-        opts: ParseOptions,
-    ) -> Result<(), adblock::lists::FilterParseError> {
-        self.0.borrow_mut().add_filter(filter, opts)
-    }
+
     fn into_content_blocking(
         &self,
     ) -> Result<(Vec<adblock::content_blocking::CbRule>, Vec<String>), ()> {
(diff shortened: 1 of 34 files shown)


js/src/lib.rs
```diff
@@ -108,27 +100,13 @@ fn filter_set_add_filters(mut cx: FunctionContext) -> JsResult<JsValue> {
         None => ParseOptions::default(),
     };
 
-    let rules: Vec<String> = json_ffi::from_js(&mut cx, rules_handle)?;
+    let rules: String = json_ffi::from_js(&mut cx, rules_handle)?;
```
antonok-edm: knowing JS, this change is going to be really annoying for downstream users to discover 😞 

If parsing as a `String` fails, can we fallback to parsing as the `Vec<String>` like previously and print a warning to the console?

src/lists.rs
```diff
@@ -221,54 +213,43 @@ impl FilterSet {
...
+            list_sources: Vec::new(),
         }
     }
 
     /// Adds the contents of an entire filter list to this `FilterSet`. Filters that cannot be
     /// parsed successfully are ignored. Returns any discovered metadata about the list of rules
     /// added.
-    pub fn add_filter_list(&mut self, filter_list: &str, opts: ParseOptions) -> FilterListMetadata {
-        self.add_filters(filter_list.lines(), opts)
+    pub fn add_filter_list(&mut self, list_text: String, opts: ParseOptions) {
```
antonok-edm: I recommend `list_text: impl Into<String>` so that we don't need `.to_string()` at every callsite where we're using a quoted `&str`.

This will still be zero-cost if the argument is already a `String`.

src/blocker.rs
```diff
@@ -451,15 +451,26 @@ impl Blocker {
...
         let mut filter_set = FilterSet::new(true);
-        filter_set.network_filters = network_filters;
-        let engine = Engine::from_filter_set(filter_set, options.enable_optimizations);
+        filter_set.add_filters(network_filters, Default::default());
+        let engine = Engine::new_with_filter_set(filter_set, options.enable_optimizations);
+        Self::from_context(engine.filter_data_context())
+    }
+
+    #[cfg(test)]
+    pub fn new_with_list_text(text: String, options: &BlockerOptions) -> Self {
```
antonok-edm, line 467: do we need both of these? at the very least they are so similar that one should wrap the other

src/engine.rs
```diff
@@ -133,6 +123,49 @@ impl Engine {
         }
     }
 
+    /// Loads rules from the given `FilterSet`. It is recommended to use a `FilterSet` when adding
+    /// rules from multiple sources.
```
antonok-edm: Since the string constructor is no longer an option, this comment can be reworded

src/engine.rs
```diff
@@ -133,6 +123,49 @@ impl Engine {
...
+    pub fn new_with_filter_set(set: FilterSet, optimize: bool) -> Self {
+        let (engine, _) = Self::from_filter_set_with_metadata(set, optimize);
+        engine
+    }
+
+    #[doc(hidden)]
+    pub fn from_filter_set_with_metadata(
```
antonok-edm: I suppose this could also be `#[cfg(test)] pub(crate)`?

src/lists.rs
```diff
@@ -185,33 +177,33 @@ pub struct FilterListMetadata {
     /// the same as the current address, meaning that it can be used to enforce the "canonical"
     /// address of the filter list.
     pub redirect: Option<String>,
+
+    /// Errors that occurred while parsing the filter list.
+    pub errors: Vec<FilterParseError>,
```
antonok-edm: hmm, `FilterListMetadata` is basically a 1:1 representation of the comments at the top of a filter list. It feels weird to add parsing errors into it. Could it be separate?

src/lists.rs
```diff
@@ -185,33 +177,33 @@ pub struct FilterListMetadata {
...
+                self.homepage = Some(value);
+            }
+            ParsedMetadata::Title(value) => {
+                self.title = Some(value);
+            }
+            ParsedMetadata::Expires(value) => {
+                self.expires = Some(value);
+            }
+            ParsedMetadata::Redirect(value) => {
+                self.redirect = Some(value);
```
antonok-edm: the `if self.field.is_none()` logic is important; existing values from earlier in the list should never be overwritten

src/lists.rs
```diff
@@ -397,6 +389,18 @@ pub enum ParsedFilter {
...
+    Homepage(String),
+    Title(String),
+    Expires(ExpiresInterval),
+    Redirect(String),
+    Unknown,
+}
+pub enum ParsedLine {
+    ParsedFilter(ParsedFilter),
+    Metadata(ParsedMetadata),
+}
```
antonok-edm: why not add the metadata variant to `ParsedFilter` directly?

src/lists.rs
```diff
@@ -510,40 +516,52 @@ pub fn parse_filter(
...
 /// Parse an entire list of filters, ignoring any errors
-pub fn parse_filters_with_metadata(
+pub fn parse_filters(
     list: impl IntoIterator<Item = impl AsRef<str>>,
     debug: bool,
     opts: ParseOptions,
-) -> (FilterListMetadata, Vec<NetworkFilter>, Vec<CosmeticFilter>) {
-    let mut metadata = FilterListMetadata::default();
-
+) -> (Vec<NetworkFilter>, Vec<CosmeticFilter>) {
```
antonok-edm, line 515: what's the reason to change this?

js/src/lib.rs
atuchin-m: A nice idea, I'll update the code.

src/lists.rs
atuchin-m: It's just about removing `parse_filters_with_metadata` in favor of `parse_filters`.
We don't use `parse_filters_with_metadata` anywhere outside.

src/lists.rs
atuchin-m: It used that option, but it turned out it doesn't force you to transfer the ownership and may result in implicit copying. 
i.e. for `s: String`, it's possible to call `set.add_filter_list(&s, default::Default())`

src/lists.rs
atuchin-m: Removed.

src/engine.rs
atuchin-m: I've made it `pub(crate)`. As for `#[cfg(test)]`: we could do it in the next PR. Currently non-test`new_with_filter_set` uses it.

src/engine.rs
atuchin-m: done.

src/blocker.rs
atuchin-m: we use the both variations a lot.
`new_with_list_text_parametrised` is a long name and the last 2 bool parameters are hard to read.
I would like to keep it as-is  at least until we rework the booleans.

js/src/lib.rs
atuchin-m: Done

atuchin-m: Done

src/lists.rs
atuchin-m: We can do that, but that will trigger more changes:
* We likely need to rename `ParsedFilter`
* `impl TryFrom<ParsedFilter> for CbRuleEquivalent` becomes less clear.
* `parse_filter_line` and `parse_filter` have to be merged.

I've tried and it looks controversial.

src/lists.rs
atuchin-m: fixed.

src/lists.rs
antonok-edm: yes, it's just for better ergonomics when e.g. working with a short hardcoded list. That could be example code, tests, or even something like Brave's localhost exception filters which are a short hardcoded string in brave-core.

atuchin-m: I'm okay to make it test-only, but for production code it leaves a room for performance issues. 
Also, tests usually use `new_with_list_text` that already supports `list_text: impl Into<String>`

src/lists.rs
antonok-edm: hmm... as this PR stands, there is currently no public way to get a `FilterListMetadata` when adding filters to a `FilterSet`? That is problematic; we will at least need to plan for it, if not in this PR

js/test/bindings.test.mjs
```diff
@@ -23,24 +23,9 @@ const { FilterSet, Engine, FilterFormat, RuleTypes, uBlockResources } =
...
-            '! Homepage: https://example.com',
-            '! Expires: 2 days',
-            '! Redirect: https://example.com/new-list.txt',
-            '||ads.com^',
-        ]);
-        assert.equal(meta.title, 'Test List');
-        assert.equal(meta.homepage, 'https://example.com');
-        assert.ok(meta.expires != null);
-        assert.equal(meta.redirect, 'https://example.com/new-list.txt');
-    });
```
antonok-edm, line 39: this is a real usecase and shouldn't be removed

antonok-edm (approved):
we need the `FilterListMetadata` part still, but otherwise OK

atuchin-m:
>hmm... as this PR stands, there is currently no public way to get a FilterListMetadata when adding filters to a FilterSet? That is problematic; we will at least need to plan for it, if not in this PR

The idea is to store the metadata in the flatbuffer in the next PR.
We can discuss API to get it in DM.

src/lists.rs
```diff
@@ -221,54 +211,45 @@ impl FilterSet {
...
         }
     }
 
     /// Adds the contents of an entire filter list to this `FilterSet`. Filters that cannot be
     /// parsed successfully are ignored. Returns any discovered metadata about the list of rules
     /// added.
-    pub fn add_filter_list(&mut self, filter_list: &str, opts: ParseOptions) -> FilterListMetadata {
-        self.add_filters(filter_list.lines(), opts)
+    pub fn add_filter_list(&mut self, list_text: String, opts: ParseOptions) -> FilterListMetadata {
+        let metadata = read_list_metadata(&list_text);
```
antonok-edm: we'll need a check here for `opts.format`.
The metadata format only applies for `FilterFormat::Standard` lists; we shouldn't try to read it for `FilterFormat::Hosts` lists.

src/engine.rs
```diff
@@ -133,6 +121,24 @@ impl Engine {
...
+        for list_source in set.list_sources {
+            for line in list_source.list_text.lines() {
+                let parsed_line = parse_filter(line, set.debug, list_source.parse_options);
+                match parsed_line {
+                    Ok(ParsedLine::Network(filter)) => network_filters.push(filter),
+                    Ok(ParsedLine::Cosmetic(filter)) => cosmetic_filters.push(filter),
+                    Err(_) => {}
+                }
+            }
+        }
```
antonok-edm, line 138: could be just `lists::parse_filters` (although I guess the intention is to replace this with a streaming implementation anyways?)

antonok-edm (approved):
feel free to merge after the check is added

src/lists.rs
atuchin-m: Done, thanks!

src/engine.rs
atuchin-m: yep, this part will be reworked.

PR #654, opened 2026-05-29
[perf] Speedup domain parsing

atuchin-m:
~~* Add a fast path in `NetworkFilter::parse` that bypasses `domain_to_ascii_cow`
for pure-ASCII hostnames (the common case);~~
* For the ASCII path, use the original string instead
of `to_string()` to avoid a redundant clone.

### DIFF
--- a/src/filters/network.rs
+++ b/src/filters/network.rs
@@ -4,6 +4,7 @@
 use memchr::memchr as find_char;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 use std::sync::LazyLock;
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


antonok-edm:
I actually looked into this previously, it was like this before. `idna` has the exact same check internally so it's not necessary to implement it ourselves.

atuchin-m:
> I actually looked into this previously, it was like this before. idna has the exact same check internally so it's not necessary to implement it ourselves.

Makes sense. We don't need a special path with `host.is_ascii()`, but we do need to avoid `clone()` when the string is reused.
I've updated the PR.

antonok-edm (approved):

PR #665, opened 2026-06-05
Store extra debug data

atuchin-m:
The PR stores more data for debugging, especially if debug = true. It also introduces source_id to distinguish  different lists added to a single engine.
It stores:
* (always) list metadata for every source ;
* (always) counters (networks filters, cosmetic filters;
* (debug mode only) source_id + line number (expect optimizable rules);
* (debug mode only) the lines from lists we failed to parse.

### DIFF
--- a/js/test/bindings.test.mjs
+++ b/js/test/bindings.test.mjs
@@ -25,17 +25,18 @@ const { FilterSet, Engine, FilterFormat, RuleTypes, uBlockResources } =
 describe('FilterSet.addFilters', () => {
       it('parses metadata comments (title, homepage, expires, redirect)', () => {
         const fs = new FilterSet();
-        const meta = fs.addFilters([
+        const { source_index, metadata } = fs.addFilters([
             '! Title: Test List',
             '! Homepage: https://example.com',
             '! Expires: 2 days',
             '! Redirect: https://example.com/new-list.txt',
             '||ads.com^',
         ].join('\n'));
-        assert.equal(meta.title, 'Test List');
-        assert.equal(meta.homepage, 'https://example.com');
-        assert.ok(meta.expires != null);
-        assert.equal(meta.redirect, 'https://example.com/new-list.txt');
+        assert.equal(metadata.title, 'Test List');
+        assert.equal(source_index, 0);
+        assert.equal(metadata.homepage, 'https://example.com');
+        assert.ok(metadata.expires != null);
+        assert.equal(metadata.redirect, 'https://example.com/new-list.txt');
     });
     it('hosts format parses IP-hostname entries', () => {
         const fs = new FilterSet();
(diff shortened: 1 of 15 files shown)


src/lists.rs
```diff
@@ -226,6 +228,8 @@ impl FilterSet {
         self.list_sources.push(ListSource {
             list_text,
             parse_options: opts,
+            metadata: metadata.clone(),
+            filename: None, // TODO: add API to pass filename
```
atuchin-m: @antonok-edm 
I suppose we need to pass filename (and maybe some extra info) to `add_filter_list` from b-c side. 
Any preferences?

antonok-edm (commented):
So, the intention behind including source location information in the filter is that `raw_line` could be removed entirely. Whenever the raw line is needed, it can be _looked up_ in the original list source.

I don't have a strong preference on the source location information being behind a debug mode or not, but my hunch is that it should be cheap enough to include unconditionally.

src/lists.rs
antonok-edm: Maybe it should just be tracked externally? There isn't always a file corresponding to a given filter list, even in brave-core (e.g. some hardcoded lists).

As long as we have a consistent index for each source, downstream users can choose on their own what data should be associated with each index.

src/network_filter_list.rs
```diff
@@ -7,35 +7,39 @@ use flatbuffers::ForwardsUOffset;
...
 use crate::flatbuffers::containers::flat_multimap::FlatMultiMapView;
 use crate::flatbuffers::unsafe_tools::fb_vector_to_slice;
 use crate::regex_manager::RegexManager;
 use crate::request::Request;
 use crate::utils::{to_short_hash, ShortHash};
 
+pub struct CheckResultDebugData {
+    pub raw_line: String,
+    pub source_index: i32,
+    pub line_number: i32,
```
antonok-edm: We may as well use `u32` for these, and reserve `u32::MAX` for "unavailable"?

js/test/bindings.test.mjs
```diff
@@ -151,7 +152,7 @@ describe('Engine.check — basic blocking', () => {
             'https://ads.example.com/t.js', 'https://pub.com', 'script', 'get', true
         );
         assert.equal(result.matched, true);
-        assert.equal(result.filter, '||ads.example.com^');
+        assert.equal(result.filter, '0:0: ||ads.example.com^');
```
antonok-edm, line 155: these really should include the raw line, source index, and line number as structured fields.

But we may have to push that to a followup. This is already becoming large and there will be a bunch of other changes involved.

atuchin-m: IMHO it's okay for now.
The good thing is that we can change it in future  without doing breaking API changes.

src/network_filter_list.rs
atuchin-m: Make sense. I've pushed the change to use u32.

js/test/bindings.test.mjs
antonok-edm: > The good thing is that we can change it in future without doing breaking API changes.

unfortunately, this is more an issue with `BlockerResult` holding `String`s. Those will need to change to a struct that has debug info fields too

Issue #391, opened 2024-08-28
support `$method` option


antonok-edm:
See [uBO wiki](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#method) for details

ShivanKaul:
@ryanbr can you give some example `$method` rules for testing?

ryanbr:
https://github.com/uBlockOrigin/uBlock-issues/wiki/Static-filter-syntax#method

Quick example of a HEAD ad: (disable shields to show)
`https://streamingcommunityz.design/`

Url: `https://lmvouuqrffurr.com/` (url may change since this post)
Blocked in uBO via 
`||com/|$xhr,3p,method=head,from=andyday.*|andyday.pe|anime3rb.com|animeflv.net|animeunity.*|animez.org|bflix.*|bflix.sh|capo5play.com|cardboardtranslation.com|cloudnestra.com|curiosidadtop.com|dlhd.*|dlhd.click|dokoembed.pw|dorawatch.net|ebook-hunter.org|embedcanais.com|embedtv.net|fbstreams.pm|fojik.*|fojik.com|freemoviesfull.com|futemais.net|giize.com|hianime.*|hoca6.com|instalker.org|kaa.mx|luckypatchers.com|luluvdoo.com|mlbb.space|moviesjoy.*|mywatchseries.*|olympustaff.com|playerflixapi.com|rivestream.org|souq-design.com|sportshub.*|spotifylists.com|streaming-community.*|streamingcommunityz.*|streamlord.com|telepisodes.org|tfp.is|tutvlive.ru|ups2up.fun|vecloud.eu|vipbox.*|vipbox.lc|watchf1full.com|way2movies.*|webxzplay.cfd|xoxocomic.com|yuppow.com`


Our current semi-work around for `method=HEAD` in https://github.com/brave/adblock-lists/blob/master/brave-unbreak.txt#L1454

```
##+js(no-fetch-if, method:HEAD)
##+js(no-xhr-if, method:HEAD)
```

* **method=post**
```
filters/privacy.txt:||lemonde.fr/*?s=$xhr,method=post
filters/privacy.txt:||perplexity.ai/rest/metrics/collect^$xhr,1p,method=post
filters/privacy.txt:||apmplus.volces.com/monitor_web/collect$xhr,method=post
filters/privacy.txt:||gator.volces.com/list$xhr,method=post
```

* **method=get**
```
[...]

PR #666, opened 2026-06-08
Support method option

boocmp:
Resolves https://github.com/brave/adblock-rust/issues/391

### DIFF
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -203,6 +203,12 @@ fn engine_check(mut cx: FunctionContext) -> JsResult<JsValue> {
                 };
                 (String::new(), debug)
             } else if let Ok(debug) = arg.downcast::<JsBoolean, _>(&mut cx) {
+                console_warn(
+                    &mut cx,
+                    "Engine.check: passing debug as the 4th boolean argument is deprecated; \
+                     pass an HTTP method string as the 4th argument (use \"\" if unspecified) \
+                     and debug as an optional 5th boolean instead.",
+                )?;
                 (String::new(), debug.value(&mut cx))
             } else if let Ok(method) = arg.downcast::<JsString, _>(&mut cx) {
                 let debug = match cx.argument_opt(5) {
--- a/js/src/lib.rs
+++ b/js/src/lib.rs
@@ -189,9 +189,9 @@ fn engine_check(mut cx: FunctionContext) -> JsResult<JsValue> {
 
     // Optional trailing args:
     // - legacy: `(debug: boolean)`
-    // - new: `(method: string | null | undefined, debug?: boolean)`
+    // - new: `(method: string, debug?: boolean)` — empty method string means unspecified
     let (method, debug) = match cx.argument_opt(4) {
-        None => (None, false),
+        None => (String::new(), false),
         Some(arg) => {
             if arg.is_a::<JsNull, _>(&mut cx) || arg.is_a::<JsUndefined, _>(&mut cx) {
                 let debug = match cx.argument_opt(5) {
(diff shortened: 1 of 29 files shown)


tests/unit/filters/network.rs
```diff
@@ -1258,4 +1267,129 @@ mod parse_tests {
             &[utils::fast_hash("some"), utils::fast_hash("primewire")]
         );
     }
+
+    #[test]
+    fn handles_method_options() {
```
atuchin-m, line 1275: Let's reduce the amount of copy-paste here by adding some constants.

tests/unit/optimizer.rs
```diff
@@ -259,41 +254,25 @@ mod optimization_tests_pattern_group {
 
             assert!(
                 filter.matches_test(
```
atuchin-m, line 256: Let's keep `matches_test` as it.
Instead add `matches_test_something` when we need to specify something meaningful.

boocmp: I didn't change matches_test, I've added parameter to Request::new

atuchin-m: Ah, I see, my bad

src/filters/network.rs
```diff
@@ -132,6 +133,14 @@ bitflags::bitflags! {
         // "Other" network request types
         const UNMATCHED = 1 << 25;
 
+        const FROM_GET = 1 << 26;
+        const FROM_HEAD = 1 << 27;
+        const FROM_POST = 1 << 30;
+
+        const FROM_METHODS = Self::FROM_GET.bits() |
```
atuchin-m: nit: FROM_ANY_METHODS?

atuchin-m (approved):

js/src/lib.rs
```diff
@@ -187,17 +187,41 @@ fn engine_check(mut cx: FunctionContext) -> JsResult<JsValue> {
     let source_url: String = cx.argument::<JsString>(2)?.value(&mut cx);
     let request_type: String = cx.argument::<JsString>(3)?.value(&mut cx);
 
-    let debug = match cx.argument_opt(4) {
+    // Optional trailing args:
+    // - legacy: `(debug: boolean)`
+    // - new: `(method: string, debug?: boolean)` — empty method string means unspecified
```
antonok-edm, line 192: let's log a deprecation warning to the console if the legacy version is used, similar to the other change we made to the JS bindings

src/engine.rs
```diff
@@ -292,7 +292,7 @@ impl Engine {
     /// `hidden_class_id_selectors` to obtain any stylesheets consisting of generic rules (if the
     /// returned `generichide` value is false).
     pub fn url_cosmetic_resources(&self, url: &str) -> UrlSpecificResources {
-        let request = if let Ok(request) = Request::new(url, url, "document") {
+        let request = if let Ok(request) = Request::new(url, url, "document", "") {
```
antonok-edm: documents will only ever be `GET` requests, so we _could_ safely specify it here

...that being said, the only time it would be relevant is if somebody uses something like this:
```
||example.com^$generichide,method=post
```
which doesn't really make sense.

Maybe `generichide` and `method` together should be considered an invalid filter?

src/request.rs
```diff
@@ -5,6 +5,13 @@ use thiserror::Error;
 use crate::url_parser;
 use crate::utils;
 
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum RequestMethod {
+    Get,
+    Head,
+    Post,
+}
```
antonok-edm, line 19: Let's include other standard method types for future proofing, along with `Other`.

Technically, an unspecified method should be treated differently from a different method we don't support yet. E.g. differentiating `DELETE` from `GET` if there are ever filters with `$method=~get|~head|~post` or `$method=get|head|post`.

antonok-edm (commented):
Thanks! couple comments and we can merge

antonok-edm (approved):
🚀

PR #668, opened 2026-06-09
Remove optimize param from Engine

atuchin-m:
The PR refactors the public interface to remove excessive `optimize` param.

### DIFF
--- a/benches/bench_cosmetic_matching.rs
+++ b/benches/bench_cosmetic_matching.rs
@@ -1,3 +1,4 @@
+use adblock::lists::FilterSet;
 use adblock::Engine;
 use criterion::*;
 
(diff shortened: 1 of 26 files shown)


benches/bench_cosmetic_matching.rs
```diff
@@ -10,7 +11,9 @@ pub fn make_engine() -> Engine {
     let rules = test_utils::rules_from_lists(["data/brave/brave-main-list.txt"]);
     let resource_json = std::fs::read_to_string("data/brave/brave-resources.json").unwrap();
     let resource_list: Vec<Resource> = serde_json::from_str(&resource_json).unwrap();
-    let mut engine = Engine::new_with_list_text_parametrised(rules, Default::default(), true, true);
```
atuchin-m, line 13: not clear why we use debug here, but it's a topic for another PR.

antonok-edm: right, I don't think debug is necessary here

js/src/lib.rs
```diff
@@ -164,19 +159,11 @@ fn engine_constructor(mut cx: FunctionContext) -> JsResult<JsBox<Engine>> {
...
-            };
-            EngineInternal::new_with_filter_set(rules, optimize)
-        }
-        None => EngineInternal::new_with_filter_set(rules, true),
-    };
+    // Legacy second argument (optimize boolean or options object) is ignored; optimization
+    // follows FilterSet.debug (!debug).
+    let _ = cx.argument_opt(1);
+
+    let engine_internal = EngineInternal::new_with_filter_set(rules);
```
antonok-edm: the optimization logic comment is outdated now

We don't necessarily need to parse the `EngineOptions` anymore, but we should at least log a warning if the extra parameter is passed

antonok-edm (approved):
LGTM apart from the comment on the JS binding

js/src/lib.rs
atuchin-m: I've added a warning.

antonok-edm (approved):

PR #678, opened 2026-06-18
Added support for parsing ABP style injections in cosmetic filters.

boocmp:
Closes https://github.com/brave/adblock-rust/issues/415

PR #685, opened 2026-06-27
Parse AdGuard `#$#` CSS injection as style/remove actions

ryanbr:
## Summary

AdGuard's `#$#` CSS-injection syntax (e.g. `businesshemden.com#$#.mnd-cookie-modal { display: none !important; }`) was rejected outright as `UnsupportedSyntax`. This routes it through the same brace-body parser added in #678 for the uBO/ABP `##selector { ... }` form, producing a `Style` (or `remove:` → `Remove`) action.

```
businesshemden.com#$#.mnd-cookie-modal { display: none !important; }
  →  Style(".mnd-cookie-modal" => "display: none !important;")
```

## Details

- The `between_sharps.starts_with('$')` branch now **consumes** the `$` and sets a flag instead of erroring. The existing `?` and leading-`@` handlers then cover the extended `#$?#` and exception `#@$#` forms with no extra code.
- A post-parse guard rejects any `#$#` rule that produced **no action**. This is the key safety check: **ABP snippet injection** also uses `#$#` but has no `{ ... }` body (e.g. `example.com#$#abort-on-property-read alert`), and must stay unsupported rather than being misparsed as a plain CSS selector.
- `#%#` / `#@%#` (AdGuard JS/scriptlet) remain unsupported — scope is limited to the CSS style/remove form.

## Behavior change

Two existing assertions in `tests/unit/lists.rs` that expected `#$#` / `#@$#` to be `UnsupportedSyntax` are updated to expect successful parsing — this is the intended contract change.

## Tests

`cargo fmt --check` clean; full lib suite passes under both default and `css-validation` features. Adds `adguard_css_injection` (style, multi-class selector, `remove:`, ABP-snippet rejection, generic rejection) and `adguard_css_injection_extended` (`#$?#`).

### DIFF
--- a/src/filters/cosmetic.rs
+++ b/src/filters/cosmetic.rs
@@ -429,6 +429,7 @@ impl CosmeticFilter {
             };
 
             let mut translate_abp_syntax = false;
+            let mut adguard_css_injection = false;
 
             // Consume filter options embedded in the `##` marker:
             let mut between_sharps = &line[after_sharp_index..second_sharp_index];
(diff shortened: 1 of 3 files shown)


ryanbr:
Superseded by #686, opened from an internal branch so the Performance CI step can post its results comment. The failure here was a fork-PR token permission issue (`pull_requests=write` not granted to fork PRs), not a regression — all benchmark ratios were ≤ 1.00.

PR #698, opened 2026-07-14
Update rust to 1.97.0

atuchin-m:
https://blog.rust-lang.org/2026/07/09/Rust-1.97.0/

### DIFF
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -438,10 +438,7 @@ impl Blocker {
 
         let mut remaining_directives = enabled_directives.difference(&disabled_directives);
 
-        let mut merged = {
-            let directive = remaining_directives.next()?;
-            String::from(*directive)
-        };
+        let mut merged = String::from(*remaining_directives.next()?);
 
         remaining_directives.for_each(|directive| {
             merged.push(',');
--- a/src/blocker.rs
+++ b/src/blocker.rs
@@ -438,10 +438,9 @@ impl Blocker {
 
         let mut remaining_directives = enabled_directives.difference(&disabled_directives);
 
-        let mut merged = if let Some(directive) = remaining_directives.next() {
+        let mut merged = {
+            let directive = remaining_directives.next()?;
             String::from(*directive)
-        } else {
-            return None;
         };
 
         remaining_directives.for_each(|directive| {
--- a/src/resources/resource_storage.rs
+++ b/src/resources/resource_storage.rs
@@ -628,7 +628,10 @@ pub(crate) fn parse_scriptlet_args(mut args: &str) -> Option<Vec<String>> {
                 args = &args[1..];
                 let i;
                 (i, needs_transform) = index_next_unescaped_separator(args, qc);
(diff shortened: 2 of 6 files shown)


boocmp (approved):

src/resources/resource_storage.rs
```diff
@@ -644,10 +647,6 @@ pub(crate) fn parse_scriptlet_args(mut args: &str) -> Option<Vec<String>> {
                         // Treating it as an error for now.
                         return None;
                     }
-                } else {
-                    // uBO pushes the entire argument, including the unmatched quote. Again, weird
-                    // and probably not intended.
-                    return None;
```
antonok-edm, line 650: would be good to keep this behavior documented somewhere

src/url_parser/parser.rs
```diff
@@ -266,13 +266,11 @@ impl<'i> Input<'i> {
     fn next_utf8(&mut self) -> Option<(char, &'i str)> {
         loop {
             let utf8 = self.chars.as_str();
-            match self.chars.next() {
-                Some(c) => {
-                    if !matches!(c, '\t' | '\n' | '\r') {
-                        return Some((c, &utf8[..c.len_utf8()]));
-                    }
+            {
```
antonok-edm: I think all of these extra scopes are unnecessary

src/blocker.rs
```diff
@@ -438,10 +438,9 @@ impl Blocker {
 
         let mut remaining_directives = enabled_directives.difference(&disabled_directives);
 
-        let mut merged = if let Some(directive) = remaining_directives.next() {
+        let mut merged = {
+            let directive = remaining_directives.next()?;
             String::from(*directive)
-        } else {
-            return None;
         };
```
antonok-edm:
```suggestion
let mut merged = String::from(*remaining_directives.next()?);
```

src/resources/resource_storage.rs
atuchin-m: it's move to `let i = i?;`

src/blocker.rs
atuchin-m: Fixed

src/url_parser/parser.rs
atuchin-m: Fixed, thanks.

src/resources/resource_storage.rs
antonok-edm: I mean the comment explaining the `i == None` case

atuchin-m: AFAIK, `i?` handles the old `return None` case. That's why the comment was moved here.

atuchin-m: I updated the comment to make it more clear.

antonok-edm (approved):

src/resources/resource_storage.rs
antonok-edm: ah sorry, that's my bad. I thought it was a different comment :sweat_smile: But the extra clarification now is good.

PR #703, opened 2026-07-16
Expose `DeserializationError`

antonok-edm:
The type is part of the public API as per `Engine::deserialize`, yet it is impossible to use it as a downstream consumer because it is not public.

These changes expose the error type and improve its usability by implementing the `thiserror::Error` trait, adding `expected`/`actual` fields to replace the `println!` logging, removing the unused `ValidationError` case, and adding documentation.

### DIFF
--- a/src/data_format/mod.rs
+++ b/src/data_format/mod.rs
@@ -28,7 +28,6 @@ pub enum DeserializationError {
     BadChecksum,
     VersionMismatch(u8),
     FlatBufferParsingError(flatbuffers::InvalidFlatbuffer),
-    ValidationError,
 }
 
 pub(crate) fn serialize_dat_file(data: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


src/data_format/mod.rs
```diff
@@ -56,13 +68,13 @@ pub(crate) fn deserialize_dat_file(serialized: &[u8]) -> Result<&[u8], Deseriali
...
-        println!(
-            "Expected hash: {:?}, actual hash: {:?}",
-            expected_hash,
-            seahash::hash(data).to_le_bytes()
-        );
-        return Err(DeserializationError::BadChecksum);
+        return Err(DeserializationError::BadChecksum {
+            // Unwrap safety: see debug_assert_eq above
+            expected: expected_hash.try_into().unwrap(),
+            actual: seahash::hash(data).to_le_bytes(),
```
atuchin-m: Let's calc hash only once

atuchin-m (approved):
LGTM with a nit

PR #706, opened 2026-07-20
Don't count comments as parse errors

atuchin-m:
The PR updates parsing logic to skip comments (instead counting them as parse errors)


antonok-edm (approved):
changed merge base to `0.14.x`

PR #710, opened 2026-07-29
Network filter lists store $domain=/$from= token buckets in a separate FlatMultiMap from pattern token buckets.

boocmp:
This splits serialized network filter token buckets into `pattern/hostname/catch-all` and `$domain=/$from=` in FlatBuffers.
Further, `$to=` filters will be stored in additional FlatMultiMap in the same way as `$from=` filters.


src/flatbuffers/fb_network_filter.fbs
```diff
@@ -40,8 +40,13 @@ table NetworkFilter {
 }
 
 table NetworkFilterList {
```
atuchin-m, line 42: looks like we need to use 2 tables instead adding extra fileds

src/network_filter_list.rs
```diff
@@ -62,34 +73,29 @@ impl NetworkFilterList<'_> {
...
+        ) {
+            return Some(result);
         }
 
-        None
+        // `filter_map` still includes single-domain-as-token keys, so source
+        // hostname hashes must be queried here as well as URL tokens.
+        self.check_tokens_in_map(
+            &self.get_filter_map(),
+            request.get_tokens_for_match(),
```
atuchin-m, line 100: can we exclude source_hostname_hashes here?

src/network_filter_list.rs
```diff
@@ -104,32 +110,113 @@ impl NetworkFilterList<'_> {
...
+    /// Invokes `on_match` for each matching filter. Returns `true` if `on_match` requested an early
+    /// stop by returning `true`.
+    fn match_filters<'a, I>(
+        &self,
+        filter_map: &FlatNetworkFilterMap<'_>,
+        tokens: I,
+        request: &Request,
+        active_tags: &HashSet<String>,
+        regex_manager: &mut RegexManager,
+        mut on_match: impl FnMut(CheckResult) -> bool,
```
atuchin-m, line 165: if we generalize early return, let's unify check/check_all

src/network_filter_list.rs
boocmp: Yes :D thanks for commit

src/flatbuffers/fb_network_filter_generated.rs
atuchin-m, line 1: The file was generated by outdated `flatc`.
Please use the version matches to `Cargo.toml`

src/flatbuffers/fb_network_filter.fbs
```diff
@@ -40,8 +40,13 @@ table NetworkFilter {
...
+  /// Pattern / hostname / catch-all token buckets.
   filter_map_index: [uint32] (required);
```
atuchin-m, line 44: nit: maybe it makes sense to rename `filter_map` to something else because now it ins't a single map that contains filters.
Let's do it a follow up.

src/network_filter_list.rs
```diff
@@ -104,32 +109,113 @@ impl NetworkFilterList<'_> {
...
-        let filter_map = self.get_filter_map();
+        self.collect_tokens_in_map(
+            &self.get_opt_domains_map(),
+            request.get_source_hostname_hashes_for_match(),
+            request,
+            active_tags,
+            regex_manager,
+            &mut filters,
+        );
+        self.collect_tokens_in_map(
```
atuchin-m: WDYT if we use `match_filters` directly instead of collect_tokens_in_map/check_tokens_in_map?

atuchin-m (approved):
LGTM with a few nits.
the generated file should be fixed before merging

src/flatbuffers/fb_network_filter_generated.rs
atuchin-m: UPD: it actually the correct diff. It was broken here: https://github.com/brave/adblock-rust/commit/a9895d33fc05f302272f7a8c1e0695d7e891234d#diff-8a328df225e8b80302f6eafa59fbedfaae34438548c37b2ac0bcd963eb08eae9

PR #715, opened 2026-07-31
Fix benchmarks in 0.14.x

atuchin-m:


.github/workflows/perf-ci.yml
```diff
@@ -27,7 +27,8 @@ jobs:
 
       - name: Run benchmarks
         run: |
-          ./.github/scripts/run-benchmarks.sh | tee -a output.txt
+          set -o pipefail
```
atuchin-m, line 30: Now the build fails if `run-benchmarks.sh` returns non-zero code.

boocmp (approved):

PR #716, opened 2026-08-03
Added $to= option matching.

### LINKED ISSUE #365: `$to` support
https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#to

boocmp:
Fix https://github.com/brave/adblock-rust/issues/365

### DIFF
--- a/src/filters/fb_network_builder.rs
+++ b/src/filters/fb_network_builder.rs
@@ -53,6 +53,7 @@ impl Default for NetworkFilterDebugData {
 struct NetworkFilterListBuilder<'a, 'f> {
     filter_map_builder: FlatMultiMapBuilder<ShortHash, NetworkFilterFlatEntry<'a>>,
     opt_domains_map_builder: FlatMultiMapBuilder<ShortHash, NetworkFilterFlatEntry<'a>>,
+    opt_to_domains_map_builder: FlatMultiMapBuilder<ShortHash, NetworkFilterFlatEntry<'a>>,
     token_frequencies: TokenSelector,
     filters_to_optimize: HashMap<ShortHash, Vec<NetworkFilter<'f>>>,
     tokens_buffer: TokensBuffer,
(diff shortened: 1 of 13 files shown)


src/filters/fb_network_builder.rs
```diff
@@ -251,15 +259,14 @@ impl<'a, 'f> NetworkRulesBuilder<'a, 'f> {
         debug_data: NetworkFilterDebugData,
         builder: &mut EngineFlatBuilder<'a>,
     ) {
-        if filter.is_badfilter() {
-            // Note: `get_id()` doesn't include BAD_FILTER bit.
-            self.bad_filter_ids.insert(filter.get_id());
+        // Exclude-only `$to=~` is rejected at parse; guard here for non-parse construction.
```
atuchin-m, line 262: Let's clarify the comment: we skip the rule only if it has only excluded domains in `$to=`.

src/filters/fb_network_builder.rs
```diff
@@ -251,15 +259,14 @@ impl<'a, 'f> NetworkRulesBuilder<'a, 'f> {
...
+        if filter.opt_to_domains.is_none() && filter.opt_to_not_domains.is_some() {
             return;
         }
 
-        // For now, filters with $to options are parsed but ignored
-        // to preserve existing matching behavior.
-        if filter.has_to_option() {
+        if filter.is_badfilter() {
```
atuchin-m, line 267: Did we reorder this intentionally?

src/filters/network.rs
```diff
@@ -911,7 +978,27 @@ impl<'a> NetworkFilter<'a> {
...
+        // Reject exclude-only `$to=~…`, and broad-only `$to=com|…` unless the filter also has
+        // a selective index key (pattern/hostname tokens or `$from=`/`$domain=`), so `$to=`
+        // stays a match-time check rather than a near-universal map key.
+        if had_to_option {
+            if filter.opt_to_domains.is_none() {
+                return Err(NetworkFilterError::ExpensiveToOption);
+            }
+            if to_only_broad {
+                let mut tokens_buffer = TokensBuffer::default();
+                match filter.get_tokens(&mut tokens_buffer) {
```
atuchin-m, line 992: I believe something wrong here. We shouldn't rely on `get_tokens` there (and it's an expensive thing).

PR #728, opened 2026-09-15
[perf] Optionally skip flatbuffer verification

atuchin-m:
The idea is to provide ability to skip flatbuffer verification for Brave Browser startup in certain scenarios:
* when we load/create a DAT file we get a token identifying the current DAT file + the engine version
* the token is stored to local_state;
* next startup we call `deserialize_with_token`. If the token matches it's safe to skip verification.


atuchin-m:
replaced by https://github.com/brave/adblock-rust/pull/730

